PR #6, opened 2020-05-19
Adding integration tests

mofterdinger:


beckermarc:
I am not sure, if we should add an integration-tests module to our standard CAP Java example. I think it just makes the app more complicated to understand. We already showcase how to write an Spring MockMVC integration test in https://github.com/SAP-samples/cloud-cap-samples-java/blob/master/srv/src/test/java/my/bookshop/CatalogServiceTest.java
@AndreasSchoesser any opinion?

mofterdinger:
I think we should also show how integration tests could look like. We already mention the Maven goal to add such an integration test module in Capire. So, why not show it in our sample application.

beckermarc:
The integration-tests module is mainly required by Cloud SDK Pipeline. From my perspective it confuses beginners more than it actually helps. So I would prefer if we show it only in conjunction with the Cloud SDK Pipeline. We should anyway think about splitting up our sample into multiple smaller samples like the Node.js colleagues have done. However that's just my opinion :)

mofterdinger:
The integration tests are working fine without the Cloud SDK Pipeline. But why not split into several projects that show different aspects of CDS / CAP development.

chgeo:
> The integration-tests module is mainly required by Cloud SDK Pipeline.

Yes.  The reason for why we created along w/ the samples and not w/ the pipeline config, is that the test code needs to match to the sample model and custom handlers.

> But why not split into several projects that show different aspects of CDS / CAP development.

We could give it a try.  The downside is that the project structure no longer reflects one-to-one the structure that people see after creating new projects.  That's something we can mitigate w/ documentation/readme.  Same than in node.js
However, I am not sure the SDK pipeline itself can handle this multi-multi-module project.  We should find out, though.

PR #20, opened 2020-07-06
Refer to a specify UI5 version in fiori.html.

mofterdinger:
- without a specific UI5 version this error occurs:

App kann nicht geöffnet werden, weil die SAP-UI5-Komponente der
Anwendung nicht geladen werden konnte.
UI5-Komponente für Navigationsabsicht "#browse-books" wurde nicht
geladen.

- Found this root cause in browser console:

Access to XMLHttpRequest at 'https://sapui5.hana.ondemand.com/resources/sap/fe/core/converters/controls/PresentationConverter.js' from origin 'http://localhost:8080' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.

- see https://ui5.sap.com/versionoverview.html to get an overview about all UI5 releases

### DIFF
--- a/app/fiori.html
+++ b/app/fiori.html
@@ -39,8 +39,8 @@
 		};
 	</script>
 
-	<script src="https://sapui5.hana.ondemand.com/test-resources/sap/ushell/bootstrap/sandbox.js"></script>
-	<script src="https://sapui5.hana.ondemand.com/resources/sap-ui-core.js"
+	<script src="https://sapui5.hana.ondemand.com/1.79.2/test-resources/sap/ushell/bootstrap/sandbox.js"></script>
+	<script src="https://sapui5.hana.ondemand.com/1.79.2/resources/sap-ui-core.js"
     	data-sap-ui-libs="sap.m, sap.ushell, sap.collaboration, sap.ui.layout"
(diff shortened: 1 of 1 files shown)


beckermarc:
This should not required. We actually intentionally run with the latest UI5 version. If such an error occurs it often helps to do a hard refresh in the browser. I got the same error the other day and that already solved the problem for me :)

mofterdinger:
> This should not be required. We actually intentionally run with the latest UI5 version. If such an error occurs it often helps to do a hard refresh in the browser. I got the same error the other day and that already solved the problem for me :)

Today a refresh solved the issue, yesterday it didn't. How can we ensure that the application will not break in the future?

beckermarc:
> Today a refresh solved the issue, yesterday it didn't. How can we ensure that the application will not break in the future?

Well for me it worked the day before yesterday :) 
We intentionally accept that the UI5 version breaks our app in order to see in that sample the latest state from the UI5 colleagues, instead of regularly having to update the UI5 version.

mofterdinger:
From my point of view this is a CAP demo application and should work all the time. It should not occasionally break, this will disappoint new users.

PR #22, opened 2020-07-10
Showcase assertion annotations

martina-gehwolf:
Annotation `@assert.format` added to the field `name` of entity `Authors`  to assert that name starts with a capital letter

### DIFF
--- a/db/schema.cds
+++ b/db/schema.cds
@@ -7,11 +7,13 @@ entity Books : cuid, managed {
   descr    : localized String(1111);
   author   : Association to Authors;
   stock    : Integer;
+  @assert.range: [0.01, 99999.00]
   price    : Decimal(9,2);
   currency : Currency;
 }
 
 entity Authors : cuid, managed {
+  @assert.format: '^\p{Lu}.*'
   name         : String(111);
   dateOfBirth  : Date;
   dateOfDeath  : Date;
--- a/db/schema.cds
+++ b/db/schema.cds
@@ -7,7 +7,6 @@ entity Books : cuid, managed {
   descr    : localized String(1111);
   author   : Association to Authors;
   stock    : Integer;
-  @assert.range: [0.01, 99999.00]
   price    : Decimal(9,2);
   currency : Currency;
 }
--- a/db/schema.cds
+++ b/db/schema.cds
@@ -13,7 +13,7 @@ entity Books : cuid, managed {
 }
 
 entity Authors : cuid, managed {
-  @assert.format: '^\p{Lu}.*'
+  @assert.format: '^\p{Lu}.*' // assert that name starts with a capital letter
   name         : String(111);
   dateOfBirth  : Date;
   dateOfDeath  : Date;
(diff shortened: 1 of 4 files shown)


srv/src/test/java/my/bookshop/AdminServiceTest.java
```diff
@@ -23,11 +28,15 @@
...
 @RunWith(SpringRunner.class)
 @SpringBootTest
+@AutoConfigureMockMvc
 public class AdminServiceTest {
 
 	@Resource(name = AdminService_.CDS_NAME)
 	private DraftService adminService;
 
+	@Autowired
+	private MockMvc mockMvc;
```
beckermarc, line 38: The idea of the tests was to show the different layers on which you can test:
1. MockMVC -> Used in CatalogServiceTest
2. local Service API -> Used in AdminServiceTest
3. Unit testing event handlers -> Used in CatalogServiceHandlerTest

I think we should not mix these. You could think about renaming CatalogServiceTest to CatalogServiceITest, as it is more of an integration test and also creating an AdminServiceITest, where you use MockMVC. Alternatively you could also write your test without MockMVC but just using the local Service API.

db/schema.cds
```diff
@@ -7,11 +7,13 @@ entity Books : cuid, managed {
   descr    : localized String(1111);
   author   : Association to Authors;
   stock    : Integer;
+  @assert.range: [0.01, 99999.00]
```
beckermarc, line 10: One could argue, why the Decimal allows 9 digits, and the validation only allows a total of up to 7 digits.

martina-gehwolf: I would actually like to remove this one, as I believe one assertion is enough to showcase the feature in general. Would you agree?

beckermarc: Yes, makes sense!

srv/src/test/java/my/bookshop/AdminServiceTest.java
martina-gehwolf: I renamed the CatalogServiceTest to CatalogServiceITest and changed my tests to fit in with the local API tests.

srv/src/test/java/my/bookshop/AdminServiceTest.java
```diff
@@ -42,4 +48,21 @@ public void testUnauthorizedAccess() {
 		adminService.newDraft(Insert.into(Orders_.class).entry(Collections.emptyMap()));
 	}
 
+	@Test(expected = ServiceException.class)
+	@WithMockUser(username = "admin")
+	public void testInvalidAuthorName() {
+		Map<String, Object> data = new HashMap<>();
```
beckermarc, line 54: You could use the typed interfaces here and in the other test method instead of the "generic" map:
```
Authors author = Authors.create()
author.setName("...");
```

martina-gehwolf: I changed to the typed interface.

PR #30, opened 2020-09-16
Prevent Fiori ID dialog popup

beckermarc:


mofterdinger (approved):
Is this annotation required by a new version of ui5 ? Or was it already missing in previous versions of this sample app?

beckermarc:
It is required by a new UI5 version.

PR #31, opened 2020-09-30
Make license files compliant to reuse tool + badges

chgeo:


beckermarc (approved):
The badge for the reuse tool still shows unregistered. Is there an additional step that we still need to do?

chgeo:
> The badge for the reuse tool still shows unregistered. Is there an additional step that we still need to do?

About to do this... It will get green only after this is merged, though.

PR #32, opened 2020-10-02
some smaller improvements

rjayasinghe:
I added some small improvements to the code and added a small docuementation snippet regarding IntelliJ usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,6 @@ Optionally, use the following steps to import the project to Eclipse:
 
 2.  In Project Explorer, change the property "Package Presentation" from "Flat" to "Hierarchical" for better understanding.
 
-
 ### Building and Running
 
 1.  To **compile** the project, right-click the file `pom.xml` in the `bookshop-parent` project root folder and select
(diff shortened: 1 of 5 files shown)


srv/src/main/java/my/bookshop/handlers/AdminServiceHandler.java
```diff
@@ -56,16 +56,17 @@
 	@Resource(name = AdminService_.CDS_NAME)
```
beckermarc, line 56: What about the bean name defined here? 
Should we move that also to the c'tor and use `@Qualifier` instead?

srv/src/main/java/my/bookshop/handlers/AdminServiceHandler.java
```diff
@@ -56,16 +56,17 @@
...
+	private final PersistenceService db;
 
-	@Autowired
-	private Messages messages;
+	private final Messages messages;
 
-	private CqnAnalyzer analyzer;
+	private final CqnAnalyzer analyzer;
+
+	public AdminServiceHandler(DraftService adminService, PersistenceService db, Messages messages, CdsModel model) {
```
beckermarc, line 65: Is the `@Autowired` annotation on the component constructor optional?

README.md
```diff
@@ -143,6 +142,22 @@ Optionally, use the following steps to import the project to Eclipse:
     Two mock users are defined for local development:
     - User: `user`, password: `user` to browse books
     - User: `admin`, password: `admin` to manage books and orders
+    
+## Using IntelliJ Idea (Community and Ultimate)
+
+IntelliJ can handle the project more or less out-of-the-box. Since some of the event handlers in the project rely on
+the code generated from the CDS model the build path of the project (module) needs to be extended
+with the folder containing the generated code. In order to add the generated code you need to add the 'gen' folder
```
beckermarc, line 150: Interesting. IntelliJ does not seem to be able to auto-detect this from the dynamic compile source directory that is added by the `cds-maven-plugin` to the Maven build. Eclipse (m2e) and VS Code (which uses m2e as well) seem to be able to detect this automatically.

srv/src/main/java/my/bookshop/handlers/AdminServiceHandler.java
rjayasinghe: exactly :)

README.md
rjayasinghe: I had to add it manually. Since IntelliJ is quite popular those days I think it's worth an addition here.

srv/src/main/java/my/bookshop/handlers/AdminServiceHandler.java
rjayasinghe: Of, I missed that one due to the unusual annotation. I'll give it a try with constructor injection.

rjayasinghe: This works with constructor injection and @Qualifier annotation. 👍

beckermarc (approved):

PR #33, opened 2020-10-07
Simplify .cdsrc.json

MarcelMerkle:


.cdsrc.json
```diff
@@ -3,50 +3,19 @@
...
     "hana": {
         "deploy-format": "hdbtable"
-    },
-    "requires": {
-        "db": {
-            "kind": "sql"
-        }
-    },
-    "odata": {
-        "version": "v4"
```
mofterdinger, line 50: Which odata version is used if it's don't specified here?

MarcelMerkle: v4 is the default according to Lothar

mofterdinger: Ok, is it suggested by the cds tools team to remove this configuration and use the default value? To a cds developer it could be not obvious which odata version is used or generated. I couldn't find any hint in the capire documentation, but maybe I missed it.

MarcelMerkle: Yes, I had a discussion with Lothar and Ecki yesterday, who suggested these changes. I agree, documentation of the .cdsrc configuration would be really helpful to customers. The only documentation that currently exists is https://github.wdf.sap.corp/pages/cap/advanced/config#build-configuration, but unfortunately it is not very detailed.

mofterdinger (approved):

.cdsrc.json
lobero: Just for clarification - from cds tools team perspective there is absolutely no concrete reason for removing the odata version setting from cds configuration. If it makes sense to have this setting from a documentation point-of-view - just leave it as it is, but technically there is no need as the default odata version has been changed to `v4` with `cds 4`. 
`cds 3` is still using odata `v2` as default.

For single-tenant apps using Classic CAP Java runtime the default still is v2 - corresponding multi-tenant scenarios have to configure the odata version otherwise v4 would be used as default as the pom.xml is not present during runtime build.
cds build is logging warnings in case some strange combination is encountered. 
Thus, there is some magic behind the scenes ;-)

lobero (approved):

PR #38, opened 2020-10-16
upgrade spring boot to 2.3.4-RELEASE

rjayasinghe:


beckermarc:
Usually we refer to the same Spring Boot version in here, which we also reference in our cds-services dependencies. This is also the Spring Boot version we tested with our latest release. For the `1.10.0` this was still `2.3.3`. We will bump the Spring Boot version to `2.3.4` for the `1.11.0` release and I'd suggest to bump those versions together here as well.
See for example https://github.com/SAP-samples/cloud-cap-samples-java/pull/36

Or is there a specific reason for updating the version (e.g. additional features)?

rjayasinghe:
No specific reason but staying up to date. 😎

rjayasinghe:
Will do it as Marc suggested 😎

PR #40, opened 2020-10-26
Execute the required cds builds in each module

MarcelMerkle:
This ensures that the required artifacts are generated before the mta build packages the respective modules

### DIFF
--- a/srv/pom.xml
+++ b/srv/pom.xml
@@ -111,7 +111,7 @@
 						</goals>
 						<configuration>
 							<commands>
-								<command>build --for java-cf --clean</command>
+								<command>build --for java-cf</command>
 
 								<command>deploy --to sqlite --dry >
 									${project.basedir}/src/main/resources/schema.sql</command>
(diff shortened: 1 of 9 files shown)


beckermarc:
Would it also be an option to run mvn clean install at the beginning of the whole MTAR build once? 
Then defining these dependencies and calling mvn in the srv build should not be required anymore.

MarcelMerkle:
@beckermarc Could you have another look?

srv/pom.xml
```diff
@@ -111,7 +111,7 @@
 						</goals>
 						<configuration>
 							<commands>
-								<command>build/all --clean</command>
+								<command>build --for java-cf --clean</command>
```
beckermarc, line 114: --clean option is not required anymore (should be the default). We should also remove that with this change.

beckermarc:
@mofterdinger what do you think about this change?

mofterdinger (approved):
looks good to me

beckermarc (approved):

beckermarc:
@MarcelMerkle Do you want to merge this PR?

PR #55, opened 2021-02-03
Fix cds build on cf and non dev environment

MarcelMerkle:
Previously the build of the mtx-sidecar and db module required a
globally installed cds-dk, which is not fulfilled for pipeline builds or
staging on CF.

### DIFF
--- a/db/package.json
+++ b/db/package.json
@@ -1,11 +1,9 @@
 {
   "name": "deploy",
   "dependencies": {
-    "@sap/hdi-deploy": "^3",
-    "@sap/cds-dk": "^3"
+    "@sap/hdi-deploy": "^3"
   },
   "scripts": {
-    "postinstall": "cd .. && db/node_modules/.bin/cds build --for hana || echo \"Skipping cds build...\"",
     "start": "node node_modules/@sap/hdi-deploy --exit"
   }
 }
--- a/db/package.json
+++ b/db/package.json
@@ -1,10 +1,11 @@
 {
   "name": "deploy",
   "dependencies": {
-    "@sap/hdi-deploy": "^3"
+    "@sap/hdi-deploy": "^3",
+    "@sap/cds-dk": "^3"
   },
   "scripts": {
-    "postinstall": "cd .. && cds build --for hana",
+    "postinstall": "cd .. && db/node_modules/.bin/cds build --for hana || echo \"Skipping cds build...\"",
     "start": "node node_modules/@sap/hdi-deploy --exit"
   }
 }
--- a/mta-multi-tenant.yaml
+++ b/mta-multi-tenant.yaml
@@ -4,6 +4,12 @@ version: 1.0.0
 description: "Multitenant Bookshop CAP Java Project with UI"
 parameters:
   enable-parallel-deployments: true
+build-parameters:
+  before-all:
+   - builder: custom
+     commands:
+      - npm install --production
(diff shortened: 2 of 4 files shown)


db/package.json
```diff
@@ -1,10 +1,11 @@
 {
   "name": "deploy",
   "dependencies": {
-    "@sap/hdi-deploy": "^3"
+    "@sap/hdi-deploy": "^3",
+    "@sap/cds-dk": "^3"
```
beckermarc, line 5: Hm, that's not really nice that we now need a `@sap/cds-dk` dependency here. Would a `devDependency` also be sufficient? This would be the better solution I guess, if required at all.

beckermarc (changes requested):

db/package.json
MarcelMerkle: I also don't like it but we need the the `cds` command here to build the module. The problem is that the mbt build tool is installing the npm modules with "--production", so a dev dependency would not be considered.

beckermarc: @chgeo what are your opinions on this?

chgeo: Don't like this approach here either.  We do have a solution for this in the mta that we generate for new Java projects.
I vote for not inventing new build mechanisms here, but just use what we offer to new projects.

```
cds add javabuildtest --add java,mta,hana,samples
```
gives you in `mta.yaml`:
```yaml
build-parameters:
  before-all:
   - builder: custom
     commands:
      - npm install --production
      - npx -p @sap/cds-dk cds build --production
```
No dependencies in `package.json`, `db/package.json` and no postinstall scripts.

beckermarc: Fully aggree that this should be consistent. We did this change https://github.com/SAP-samples/cloud-cap-samples-java/pull/40 as part of an experiment we discussed with Lothar, but if it creates these follow-up problems, we should probably rather revert back to the previous approach / the one we promote currently with  `cds init`.

beckermarc: The current approach promoted by `cds init` has the drawback that the cds build is executed twice in case of an MTAR build.

MarcelMerkle: Only the java part should be executed twice.

chgeo: Right, this could be improved.  But rather should do that w/ Lothar and Markus instead of having yet another custom solution to the problem.
Double build time is not an issue here.

beckermarc: Yes fully agree. @MarcelMerkle can you revert the changes and implement the solution based on the way `cds init` currently behaves. That the Java module only builds the srv part can be kept as a continued experiment I guess. If we like how that feels we can also update that in our Maven archetype.

MarcelMerkle: I really like `npx -p @sap/cds-dk cds build --production`! Didn't know that this was possible.

MarcelMerkle: To optimize the build time, we could think of specifying only the necessary build tasks for the global cds-build. But for now I would keep it consistent with `cds init`.

MarcelMerkle: @chgeo What is the reason for the `npm install --production` step? The `package.json` in the root directory has no dependencies.

beckermarc (approved):
LGTM: Depending on Christians answer, we could also remove `npm install --production` from my perspective. It might be required if apps have dependencies to other cds models in their package.json.

MarcelMerkle:
I will merge this PR to fix the issue. If we want we can remove the `npm install` step later on.

PR #58, opened 2021-02-05
Enhance bookshop with S/4 extension scenario

beckermarc:
Added:
- S/4 remote OData extension scenario incl. using messaging for keeping replicas up-to-date.
- Runs fully mocked locally, incl. mocked S/4 API + messaging capabilities.
- Runs in the cloud by leveraging S/4 BusinessPartner API from SAP API Hub (API Key needs to be provided)

Notes:
- Messaging configuration for the cloud is **not** provided by default, as SAP API Hub does not emit events. However the app also works without it. Technically it would be possible to bind a destination service for an actual S/4 system + enterprise-messaging service to the app and everything would work out-of-the-box, only based on deployment configuration changes.


app/orders/fiori-service.cds
```diff
@@ -24,7 +24,32 @@ annotate AdminService.OrderItems with {
...
+					{ $Type: 'Common.ValueListParameterDisplayOnly', ValueListProperty: 'Country'},
+					{ $Type: 'Common.ValueListParameterDisplayOnly', ValueListProperty: 'StreetName'},
+					{ $Type: 'Common.ValueListParameterDisplayOnly', ValueListProperty: 'HouseNumber'},
+				]
+			}
+		}
+	);
+}
 
+////////////////////////////////////////////////////////////////////////////
```
AndreasSchoesser, line 49: Should that section comment be before line 27? There, the UI annoations seem to begin IMHO

AndreasSchoesser (approved):

s4-api-mock.http
```diff
@@ -0,0 +1,4 @@
+PATCH http://localhost:8080/api/API_BUSINESS_PARTNER/A_BusinessPartnerAddress(BusinessPartner='10401010',AddressID='200')
```
MarcelMerkle, line 1: What is this file for?

srv/external.cds
```diff
@@ -0,0 +1,46 @@
...
+    @readonly PostalCode,
+    @readonly StreetName,
+    @readonly HouseNumber
+  }
+
+  /**
+   * Add asynchronous eventing API
+   */
+  event BO_BusinessPartner_Changed {
+    ![KEY]: Array of {
```
MarcelMerkle, line 25: How does this `![KEY]` notation work?

app/orders/fiori-service.cds
beckermarc: The comment should have contrasted the other comment `Common` in line 12. As the block from 27 to 47 also added some Common annotations to elements of the entities I added it there.

The annotations starting in line 53 describe how the List / Object page UIs should look like, thats why I titled them with `UI`. Couldn't come up with a better name 😕

s4-api-mock.http
beckermarc: This file provides an example request that can be executed using the VS Code Rest Client plugin (https://marketplace.visualstudio.com/items?itemName=humao.rest-client). It simulates a change of the address on the mocked Business Partner API and triggers the mocked `BO.BusinessPartner.Changed` event, defined in `ApiBusinessPartnerEventMockHandler`. This in turn allows to mock the event receiving in a local scenario to test the corresponding event handler defined in `AdminServiceAddressHandler`.

srv/external.cds
beckermarc: the `![]` notation is the cds notation for quoting. As `key` is a cds keyword it needs to be quoted to be used as an element name here.

beckermarc: I didn't really find this documented on capire, but there are some hints scattered across release notes: https://github.wdf.sap.corp/pages/cap/releases/nov20#quickfixes-for-deprecated-identifiers. Also writing it differently (e.g. using double quotes) creates a compiler warning.

srv/src/main/java/my/bookshop/handlers/AdminServiceAddressHandler.java
```diff
@@ -0,0 +1,189 @@
...
+public class AdminServiceAddressHandler implements EventHandler {
+
+	private final static Logger logger = LoggerFactory.getLogger(AdminServiceAddressHandler.class);
+
+	// We are mashing up the AdminService with two other services...
+	@Autowired
+	private PersistenceService db;
+
+	@Resource(name = ApiBusinessPartner_.CDS_NAME)
+	private CdsService bupa;
```
MarcelMerkle, line 69: Should we inject here through the constructor as Robin already suggested for our other handlers?

srv/src/main/java/my/bookshop/handlers/AdminServiceAddressHandler.java
```diff
@@ -0,0 +1,189 @@
...
+		}
+
+		// forward the selected columns (in Java, we cannot use our simplified projection Addresses yet)
+		List<CqnSelectListItem> columns = context.getCqn().items();
+		boolean hasStar = columns.stream().anyMatch(CqnSelectListItem::isStar);
+		if (hasStar || columns.isEmpty()) {
+			select.columns(getRelevantColumns());
+		} else {
+			List<String> relevantColumnsList = Arrays.asList(getRelevantColumns());
+			select.columns(columns.stream().filter(i -> relevantColumnsList.contains(i.asValue().value().asRef().firstSegment())).collect(Collectors.toList()));
```
MarcelMerkle, line 103: I'm not sure if we can be sure that `i` is a reference here, but I guess if don't assume this, it would make this logic even more complex..

srv/src/main/java/my/bookshop/handlers/AdminServiceAddressHandler.java
```diff
@@ -0,0 +1,189 @@
...
+			String addressId = order.getShippingAddressAddressID();
+			if(addressId != null) {
+				Result replica = db.run(Select.from(addresses).where(a -> a.BusinessPartner().eq(businessPartner).and(a.AddressID().eq(addressId))));
+				// check if the address was not yet replicated
+				if(replica.rowCount() < 1) {
+					Result remoteAddresses = bupa.run(Select.from(externalAddresses).columns(getRelevantColumns())
+							.where(a -> a.BusinessPartner().eq(businessPartner).and(a.AddressID().eq(addressId))));
+
+					if(remoteAddresses.rowCount() == 1) {
+						db.run(Upsert.into(addresses).entries(remoteAddresses));
```
MarcelMerkle, line 137: Do we want to advertise `Upsert` here? I thought, ideally we want to get rid of this.

srv/src/main/java/my/bookshop/handlers/AdminServiceAddressHandler.java
beckermarc: Yes I think this is something we can ignore for now. In the future this code will be completely removed, as soon as we support projections on remote entities. This here is just a poor mans solution for implementing this feature.

srv/src/main/java/my/bookshop/handlers/AdminServiceAddressHandler.java
beckermarc: For these replication scenarios I think upsert is the ideal statement. It ensures that the replicated data is _**replaced**_ with the latest version from the remote system. So far we only had issues with `Upsert` on OData API level and with managed entities, but here in that case none of these two is the case.

srv/src/main/java/my/bookshop/handlers/AdminServiceAddressHandler.java
beckermarc: yes, I can change it.

srv/src/main/java/my/bookshop/handlers/AdminServiceAddressHandler.java
MarcelMerkle: But here we explicitly check if the `replica` already exists. So in my opinion we could always use an `Insert` here. 

I'm also thinking about if it would make sense to do the insert/upsert in a separate transaction, because otherwise a parallel transaction could also insert the `remoteAddresses` and this transaction would fail when it is commited.

beckermarc: You are right, in my previous comment I was thinking about the usage of Upsert in the `updateBusinessPartnerAddresses` method.

However from my perspective the `Upsert` statement expresses the idempotency that you are referring to. I'd like to avoid to have to start a separate transaction here and would hope that this "last one wins" is ensured by the `Upsert` statement.

MarcelMerkle: I don't think that the `Upsert` statement can guarantee this without opening a separate transaction, which is not the case I think. (Please correct me if I'm wrong @agoerler) If the `remoteAddress` does not exist when the transaction is opened, the conflict will only become visible when the transaction is committed.

beckermarc: True, however with SQLite we are currently limited to a single connection, and can't open nested transactions here at the moment. We should revisit this, once we switch to H2 :) I will add a comment about this to the code.

PR #59, opened 2021-02-08
Replace deprecated property `spring.resources.static-locations`

mofterdinger:
with `spring.web.resources.static-locations` in application.yaml

- also see:
https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/web/ResourceProperties.html#getStaticLocations--

### DIFF
--- a/srv/src/main/resources/application.yaml
+++ b/srv/src/main/resources/application.yaml
@@ -1,8 +1,6 @@
 ---
 spring:
-  web:
-    resources:
-      static-locations: "file:./app"
+  web.resources.static-locations: "file:./app"
 cds:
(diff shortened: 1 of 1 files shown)


srv/src/main/resources/application.yaml
```diff
@@ -1,6 +1,8 @@
 ---
 spring:
-  resources.static-locations: "file:./app"
+  web:
```
beckermarc, line 3: Thanks. Can you keep the single-line definition to make the configuration file a little more compact?

beckermarc (changes requested):

beckermarc (approved):

PR #63, opened 2021-03-04
Bookshop extension

davidhunglam:
Enhanced bookshop with book reviews feature:
- Added schema for Reviews
- Added Genres to Books
- Added book rating service
- Added Reviews facet to book object page
- Extended book object page with additional facets
- Added dedicated review service
- Added review management section to Fiori UI
- Added example Vue app for browsing books and reviews
- Added some tests
- Extended README.md


AndreasSchoesser:
Hi, 

in the readme, could you add the screenshot from the review service instead with a view reviewed books and some "stars"? Would IMHO look much more impressive :)

Thx and best regards,
Andreas

AndreasSchoesser:
All right, at least it's consistent with the other views :)

AndreasSchoesser:
What about the stars that are displayed in the reviews overview => It seems like you could alter this by editing with the mouse visually, but it doesn't have any effect. Is it possible to disable the useless editing? Even cooler would be if, when you create a review, you could select the stars with the mouse instead of entering a score number...

AndreasSchoesser:
When I try to edit someone else's review I get the error: "Entity 'ReviewService.Reviews' with key(s) 'IsActiveEntity=true,ID=8298deea-443c-372e-bba3-628b003f6845' not found" instead of "not authorized" or something like this...

beckermarc:
> When I try to edit someone else's review I get the error: "Entity 'ReviewService.Reviews' with key(s) 'IsActiveEntity=true,ID=8298deea-443c-372e-bba3-628b003f6845' not found" instead of "not authorized" or something like this...

This is the expected behaviour because we currently do not distinguish between 404 not found and 403 not authorized on instance-based authorizations. Both result in an UPDATE/DELETE that doesn't hit any entity. This behaviour would be changed by the BLI we discussed today about checking authorizations on modifications using additional Selects.

davidhunglam:
> What about the stars that are displayed in the reviews overview => It seems like you could alter this by editing with the mouse visually, but it doesn't have any effect. Is it possible to disable the useless editing? Even cooler would be if, when you create a review, you could select the stars with the mouse instead of entering a score number...

Yes, that is another thing I could not find out how to achieve easily. I tried to use field control to mark it as readonly, but unfortunately to no effect. If you check the console it even sends a request, but as reviews are exposed as readonly in the CatalogService this is forbidden and thus seems to have no effect.

I would love to use the stars in the editing view of the AddReview action, but that's also something I've not been able to figure out yet, unfortunately...

davidhunglam:
> What about the stars that are displayed in the reviews overview => It seems like you could alter this by editing with the mouse visually, but it doesn't have any effect. Is it possible to disable the useless editing? Even cooler would be if, when you create a review, you could select the stars with the mouse instead of entering a score number...

Yes, that is another thing I could not find out how to achieve easily. I tried to use field control to mark it as readonly, but unfortunately to no effect. If you check the console it even sends a request, but as reviews are exposed as readonly in the CatalogService this is forbidden and thus seems to have no effect.

I would love to use the stars in the editing view of the AddReview action, but that's also something I've not been able to figure out yet, unfortunately...

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -2,56 +2,223 @@
...
+	 * @param context {@link ReviewContext}
+	 */
+	@On(entity = Books_.CDS_NAME)
+	public void onAddReview(AddReviewContext context) {
+		Integer rating = context.getRating();
+		String title = context.getTitle();
+		String text = context.getText();
+
+		String username = context.getUserInfo().getName();
+		String bookId = (String) analyzer.analyze(context.getCqn()).targetKeys().get(Books.ID);
```
MarcelMerkle, line 104: In `validateBook` we anyway do a db request to check if the book exists. You can use the cqn statement from the context for that and from the resulting book you could easily get the book ID. Then you don't have to use cqn analyzer, because right now you extract the book ID to later construct the exact same statement that you already have here in the context.

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -2,56 +2,223 @@
...
 		}
 	}
+
+	private void validateBook(Reviews review) {
+		if (review.getBookId() == null) {
+			throw new ServiceException(ErrorStatuses.BAD_REQUEST, MessageKeys.BOOK_MISSING)
+					.messageTarget(Reviews_.class, r -> r.book_ID());
+		}
+
+		Optional<Row> row = db.run(Select.from(CatalogService_.BOOKS).byId(review.getBookId())).first();
```
MarcelMerkle, line 218: Why do you use the `PersistenceService` and not the `CatalogService` here?

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -2,56 +2,223 @@
...
+		review.setBookId(bookId);
+		review.setReviewer(username);
+		review.setRating(rating);
+		review.setTitle(title);
+		review.setText(text);
+
+		validateReview(review);
+
+		Result res = db.run(Insert.into(CatalogService_.REVIEWS).entry(review));
+		Optional<Reviews> inserted = res.first(Reviews.class);
```
MarcelMerkle, line 116:
```suggestion
		Reviews inserted = res.single(Reviews.class);
```

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -2,56 +2,223 @@
...
+	}
+
+	/**
+	 * Recalculates and sets the book rating after a new review for the given book.
+	 * 
+	 * @param context {@link ReviewContext}
+	 */
+	@After(entity = Books_.CDS_NAME)
+	public void afterAddReview(AddReviewContext context) {
+		String bookId = (String) analyzer.analyze(context.getCqn()).targetKeys().get(Books.ID);
```
MarcelMerkle, line 129: The `bookID` should also be contained in the result.

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -2,56 +2,223 @@
...
+			loadStockIfNotSet(b);
+			discountBooksWithMoreThan111Stock(b);
+		});
+	}
+
+	@After
+	public void setIsReviewable(CdsReadEventContext context, List<Books> books) {
+		String user = context.getUserInfo().getName();
+
+		Result result = db
```
MarcelMerkle, line 145: Could be improved if we only filter for the books that are acutally in the list, but I think for the demo case it is fine. Although maybe this could be a nice example for parameterized statements.

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -2,56 +2,223 @@
...
+	}
+
+	private void validateReview(Reviews review) {
+		validateBook(review);
+		validateRating(review);
+	}
+
+	private void validateRating(Reviews review) {
+		Integer rating = review.getRating();
+		if (rating == null) {
```
MarcelMerkle, line 201: Is this a valid scenario?

srv/src/main/java/my/bookshop/handlers/ReviewServiceHandler.java
```diff
@@ -0,0 +1,61 @@
...
+public class ReviewServiceHandler implements EventHandler {
+
+	private CdsService catalogService;
+
+	ReviewServiceHandler(@Qualifier(CatalogService_.CDS_NAME) CdsService catalogService) {
+		this.catalogService = catalogService;
+	}
+
+	@Before(event = { CdsService.EVENT_CREATE, CdsService.EVENT_UPDATE })
+	public void beforeAddReview(Stream<Reviews> reviews) {
```
MarcelMerkle, line 35: Do we need this handler if the entity is readonly?

srv/src/test/java/my/bookshop/handlers/CatalogServiceHandlerTest.java
```diff
@@ -29,11 +57,112 @@ public void testDiscountHandler() {
...
+					testCase.review.title, testCase.review.text);
+			assertThrows(ServiceException.class, () -> catalogService.emit(context), testCase.exceptionMessage);
+
+		});
+	}
+
+	/*
+	 * Holder class for book review information.
+	 */
+	private class BookReview {
```
MarcelMerkle, line 132: You could also use the generated POJO interfaces for this.

srv/src/test/java/my/bookshop/handlers/CatalogServiceHandlerTest.java
```diff
@@ -29,11 +57,112 @@ public void testDiscountHandler() {
...
+						exMessage2));
+
+		testCases.forEach(testCase -> {
+			AddReviewContext context = addReviewContext(testCase.review.bookId, testCase.review.rating,
+					testCase.review.title, testCase.review.text);
+			assertThrows(ServiceException.class, () -> catalogService.emit(context), testCase.exceptionMessage);
+
+		});
+	}
+
```
MarcelMerkle, line 128: Should we also add a test to check if a user cannot create more than one review?

MarcelMerkle (changes requested):

srv/admin-service.cds
```diff
@@ -1,28 +1,45 @@
...
+annotate AdminService.Books with @cds.search :
+{
+  descr,
+  title
+};
 
 // Enable Fiori Draft for Orders
 annotate AdminService.Orders with @odata.draft.enabled;
-annotate AdminService.Books with @odata.draft.enabled @Capabilities.InsertRestrictions.Insertable: false;
+annotate AdminService.Books with @odata.draft.enabled;
```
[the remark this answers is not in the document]
davidhunglam: This annotation prevents to add new books to the bookshop which I think the admin should be able to.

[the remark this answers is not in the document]
beckermarc: We added this intentionally to show that draft also works with update only entities. But yes from a business perspective this might not be the most reasonable thing..

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
davidhunglam: Primarily, because the `PersistenceService` injection existed before already and I decided to reuse it instead of injecting the `CatalogService` in addition. Is it preferred to use the `CatalogService`?

srv/src/main/java/my/bookshop/handlers/ReviewServiceHandler.java
davidhunglam: In the ReviewService the `Reviews` entity is not readonly, only within the CatalogService.

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
davidhunglam: When using the Fiori UI this cannot happen, as rating will always be set. But, if sending the request via Postman for example while not setting the rating in your payload this can potentially happen.

srv/src/main/java/my/bookshop/handlers/ReviewServiceHandler.java
MarcelMerkle: Yes, right. But why don't we use the `ReviewService` then in the `CatalogServiceHandler` to add the review?

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
MarcelMerkle: But should we not rather throw an exception in this case instead of generating a random rating?

davidhunglam: Yeah we could, I just implemented it in the same way it is in the nodejs sample.

MarcelMerkle: Ok, I wonder what the motivation was. @David-Kunz Could you comment?

srv/src/main/java/my/bookshop/handlers/ReviewServiceHandler.java
davidhunglam: Well, yeah, the `ReviewService` implementation came after I've implemented the `CatalogServiceHandler` and the `AddReview` action in there. The separate `ReviewService` was thought of as an experiment to try more customization options in the Fiori Elements UI, as they seem to be quite restrictive for bound actions. And in the end I left it, as it demonstrates some other things quite nicely.

And in the `CatalogService` the `PersistenceService` was already injected, so I just reused it.

But if you prefer I can inject the `ReviewService` and use that one instead. That way I can remove the Review validation logic in there, as it's already handled by the `ReviewService`.

mta-single-tenant.yaml
```diff
@@ -85,5 +85,5 @@ resources:
   - name: bookshop-java-public-hdi-container
     type: org.cloudfoundry.managed-service
     parameters:
-      service: hanatrial
+      service: hana
```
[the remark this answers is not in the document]
davidhunglam: I'll revert that.

srv/src/main/java/my/bookshop/handlers/AdminServiceHandler.java
```diff
@@ -69,9 +70,10 @@
 	}
 
 	/**
-	 * Validate correctness of an order before finishing the order proces:
-	 * 1. Check Order amount for each Item and return a message if amount is empty or <= 0
-	 * 2. Check Order amount for each Item is available, return message if the stock is too low
+	 * Validate correctness of an order before finishing the order proces: 1. Check
```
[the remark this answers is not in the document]
davidhunglam: I'll revert that.

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
David-Kunz: @MarcelMerkle , I didn't create this handler, but from a 'business scenario' it doesn't make any sense to generate a random rating. I guess it was just done for easy 'generation' of many ratings, which can come handy in demos.

srv/src/main/java/my/bookshop/handlers/ReviewServiceHandler.java
MarcelMerkle: Yeah, that's what I also thought. We wouldn't duplicate the validation logic. Also later on, we could maybe even separate the two services in separate micro-services if we want to showcase that.

srv/src/main/java/my/bookshop/BookRatingService.java
```diff
@@ -0,0 +1,63 @@
...
+		}
+	}
+
+	/**
+	 * Sets the average rating for the given book.
+	 *
+	 * @param bookId
+	 */
+	public void setBookRating(String bookId) {
+		Result run = db.run(Select.from(Bookshop_.REVIEWS).where(r -> r.book_ID().eq(bookId)));
```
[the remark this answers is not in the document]
MarcelMerkle: No, in general not. Here I used the filter to navigate to the `Reviews` via the `Books`. This way you don't have to use the generated foreign key `book_ID`.

srv/src/test/java/my/bookshop/handlers/CatalogServiceHandlerTest.java
MarcelMerkle: It's also fine like this. Would just save some codelines

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -193,26 +197,20 @@ private void loadStockIfNotSet(Books b) {
...
-	}
-
-	private String validateBook(AddReviewContext context) {
-		try {
-			Books book = db.run(context.getCqn()).single(Books.class);
-			return book.getId();
-		} catch (EmptyResultException ex) {
-			throw new ServiceException(ErrorStatuses.BAD_REQUEST, MessageKeys.BOOK_MISSING);
-		}
+	private Reviews toCatalogServiceReviews(cds.gen.reviewservice.Reviews review) {
```
MarcelMerkle, line 200: I think since we don't rename any field, `Struct.access(review).as(Reviews.class)` should be enough.

MarcelMerkle (approved):
LGTM. @beckermarc Do you also want to have a look?

beckermarc:
Yes, I'd like to take a look as well, however didn't find the time yet due to my DoD duties this week :/

MarcelMerkle:
Yeah, I already suspected that. No problem, then just review it next week :)

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -2,56 +2,201 @@
...
+	public void beforeAddReview(AddReviewContext context) {
+		String reviewer = context.getUserInfo().getName();
+		String bookId = (String) analyzer.analyze(context.getCqn()).targetKeys().get(Books.ID);
+
+		Result result = db.run(Select.from(CatalogService_.REVIEWS)
+				.where(review -> review.book_ID().eq(bookId).and(review.reviewer().eq(reviewer))));
+
+		if (result.first().isPresent()) {
+			throw new ServiceException(ErrorStatuses.METHOD_NOT_ALLOWED,
+					"User not allowed to add more than one review for a given book").messageTarget(Reviews_.class,
```
beckermarc, line 93: Can we add this error message text to our resource bundles and provide a german translation?

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -2,56 +2,201 @@
...
 			b.setTitle(String.format("%s -- 11%% discount", b.getTitle()));
 		}
 	}
 
 	private void loadStockIfNotSet(Books b) {
-		if (b.getStock() == null) {
-			b.setStock(db.run(Select.from(BOOKS).byId(b.getId()).columns(Books_::stock)).single(Books.class).getStock());
+		if (b.getId() != null && b.getStock() == null) {
+			b.setStock(
+					db.run(Select.from(BOOKS).byId(b.getId()).columns(Books_::stock)).single(Books.class).getStock());
```
beckermarc, line 198: Can we avoid this reformatting?

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -2,56 +2,201 @@
...
+		});
+	}
+
+	@After
+	public void setIsReviewable(CdsReadEventContext context, List<Books> books) {
+
+		String user = context.getUserInfo().getName();
+		List<String> bookIds = books.stream().map(b -> b.getId()).collect(Collectors.toList());
+
+		Select<Reviews_> query = Select.from(CatalogService_.BOOKS, b -> b.filter(b.ID().in(bookIds)).reviews())
```
beckermarc, line 151:
```suggestion
		CqnSelect query = Select.from(CatalogService_.BOOKS, b -> b.filter(b.ID().in(bookIds)).reviews())
```

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -2,56 +2,201 @@
...
+	public void onSubmitOrder(SubmitOrderContext context) {
+		Integer amount = context.getAmount();
+		String bookId = context.getBook();
+
+		Optional<Books> book = db.run(Select.from(BOOKS).columns(Books_::stock).byId(bookId)).first(Books.class);
+
+		book.orElseThrow(() -> new ServiceException(ErrorStatuses.NOT_FOUND, MessageKeys.BOOK_MISSING)
+				.messageTarget(Books_.class, b -> b.ID()));
+
+		Integer stock = book.map(Books::getStock).get();
```
beckermarc, line 174:
```suggestion
		int stock = book.map(Books::getStock).get();
```

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -2,56 +2,201 @@
...
+
+		if (stock >= amount) {
+			db.run(Update.entity(BOOKS).byId(bookId).data(Books.STOCK, stock -= amount));
+
+			ReturnCatalogServiceSubmitOrder result = ReturnCatalogServiceSubmitOrder.create();
+			result.setStock(stock);
+
+			context.setResult(result);
+			context.setCompleted();
```
beckermarc, line 183: `setCompleted` is automatically triggered by `setResult`

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -2,56 +2,201 @@
...
+		} else {
+			throw new ServiceException(ErrorStatuses.CONFLICT, String.format("%d exceeds stock for book", amount));
```
beckermarc, line 185: Please use the built-in formatting capabilities of `ServiceException` and also consider adding the text to the resource bundles and providing a german translation.

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -2,56 +2,201 @@
...
+		for (Books book : books) {
+			if (reviewedBooks.contains(book.getId())) {
+				book.setIsReviewable(false);
 			}
-		);
+		}
+	}
+
+	@On
+	public void onSubmitOrder(SubmitOrderContext context) {
```
beckermarc, line 165: Should this action also create an order for the user?

srv/src/main/java/my/bookshop/handlers/ReviewServiceHandler.java
```diff
@@ -0,0 +1,60 @@
...
+		});
+	}
+
+	private void validateRating(Reviews review) {
+		Integer rating = review.getRating();
+		if (rating < 1 || rating > 5) {
+			throw new ServiceException(ErrorStatuses.BAD_REQUEST, MessageKeys.REVIEW_INVALID_RATING)
+			.messageTarget(Reviews_.class, r -> r.rating());
+		}
+	}
```
beckermarc, line 48: Is this ever triggered? Shouldn't this be covered by the `@assert.enum` annotation on the `Reviews` entity?

beckermarc (changes requested):
Looks very good overall. Sorry for the many comments, but I think especially in our samples we need to take care to be concise and clear and easy to understand :)

srv/review-service.cds
```diff
@@ -0,0 +1,50 @@
...
+    // input validation
+    annotate Reviews with {
+        subject @mandatory;
+        title   @mandatory;
+        rating  @assert.enum;
+    }
+
+    // access control restrictions
+    annotate ReviewService.Reviews with @restrict : [
+    {
```
beckermarc, line 25: Why not change it that the authenticated-user can only read its own reviews and the admin can read all?
We could do it something like that: `grant * to authenticated-user where reviewer = $user OR grant * to admin`

app/common.cds
```diff
@@ -1,98 +1,464 @@
 /*
   Common Annotations shared by all apps
 */
-
-using { my.bookshop as my } from '../db/schema';
+using
```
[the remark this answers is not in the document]
davidhunglam: I've been using the Eclipse CDS file formatter... I'll use the one in VSCode with the default setting and reformat all CDS files that I changed... hopefully it will look fine, then.

app/common.cds
```diff
@@ -1,98 +1,464 @@
...
-	UI: {
-  	HeaderInfo: {
-  		TypeName: '{i18n>Book}',
-  		TypeNamePlural: '{i18n>Books}',
-  		Title: {Value: title},
-  		Description: {Value: author.name}
-  	},
-	}
+annotate my.Books with
+@(UI :
```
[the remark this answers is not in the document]
davidhunglam: see comment above

app/vue/index.html
```diff
@@ -0,0 +1,104 @@
+<!DOCTYPE html>
```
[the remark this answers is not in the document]
davidhunglam: This technically replaces it, yes. I'll remove the old one.

db/books.cds
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Hierarchically organized Code List for Genres
+ */
+entity Genres : sap.common.CodeList
+{
+    key ID : Integer;
+        parent : Association to Genres;
+        children : Composition of many Genres
+                       on children.parent = $self;
```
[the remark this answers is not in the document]
davidhunglam: see comment about formatting above

db/books.cds
```diff
@@ -0,0 +1,49 @@
+namespace my.bookshop;
+
+using
+{
```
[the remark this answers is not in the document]
davidhunglam: see comment about formatting above

db/reviews.cds
```diff
@@ -0,0 +1,39 @@
...
+    managed,
+    cuid
+} from '@sap/cds/common';
+
+entity Reviews : cuid, managed
+{
+    @cds.odata.ValueList
+    book : Association to my.Books;
+    @Core.Computed
+    reviewer : User;
```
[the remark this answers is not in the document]
davidhunglam: Yeah, I guess the reviewer is redundant and we can use the createdBy or modifiedBy User instead.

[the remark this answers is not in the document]
davidhunglam: Yeah, I guess the reviewer is redundant and we can use the createdBy or modifiedBy User instead.

db/reviews.cds
```diff
@@ -0,0 +1,39 @@
...
+    rating : Rating;
+    title : String(111);
+    text : String(1111);
+    @Core.Computed
+    date : DateTime;
```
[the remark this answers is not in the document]
davidhunglam: Same as with the reviewer I guess, we can drop date and use modifiedAt instead.

srv/cat-service.cds
```diff
@@ -1,10 +1,65 @@
...
+    },
+    {
+        grant : 'DELETE',
+        to    : 'admin'
+    },
+    {
+        grant : 'DELETE',
+        to    : 'authenticated-user',
+        where : 'reviewer=$user'
+    }
```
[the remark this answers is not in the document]
davidhunglam: You're right, I'll remove these.

srv/review-service.cds
```diff
@@ -0,0 +1,50 @@
...
+
+    @readonly
+    entity Authors as projection on my.Authors;
+
+    // input validation
+    annotate Reviews with {
+        subject @mandatory;
+        title   @mandatory;
+        rating  @assert.enum;
+    }
```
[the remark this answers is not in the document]
davidhunglam: Will move these lines to the `reviews.cds` then.

srv/src/main/java/my/bookshop/Application.java
```diff
@@ -10,4 +12,12 @@ public static void main(String[] args) {
 		SpringApplication.run(Application.class, args);
 	}
 
+	@Bean
+	public CharacterEncodingFilter characterEncodingFilter() {
```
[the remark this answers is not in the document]
davidhunglam: Without this, in the vue app you don't properly see the rating stars.

Screenshot 2021-03-17 at 18 27 13

srv/src/main/java/my/bookshop/BookRatingService.java
```diff
@@ -0,0 +1,63 @@
...
+import cds.gen.my.bookshop.Books;
+import cds.gen.my.bookshop.Bookshop_;
+import cds.gen.my.bookshop.Reviews;
+
+/**
+ * Takes care of calculating the average rating of a book based on its review
+ * ratings.
+ */
+@Component
+public class BookRatingService {
```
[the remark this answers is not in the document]
davidhunglam: Okay, I'll rename it.

srv/src/main/java/my/bookshop/handlers/AdminServiceAddressHandler.java
```diff
@@ -51,7 +51,8 @@
 import my.bookshop.MessageKeys;
 
 /**
- * Custom handler for the Admin Service Addresses, which come from a remote S/4 System
+ * Custom handler for the Admin Service Addresses, which come from a remote S/4
+ * System
```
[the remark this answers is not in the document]
davidhunglam: I'll revert it.

srv/src/main/java/my/bookshop/handlers/AdminServiceHandler.java
```diff
@@ -59,7 +59,8 @@
 
 	private final CqnAnalyzer analyzer;
 
-	AdminServiceHandler(@Qualifier(AdminService_.CDS_NAME) DraftService adminService, PersistenceService db, Messages messages, CdsModel model) {
+	AdminServiceHandler(@Qualifier(AdminService_.CDS_NAME) DraftService adminService, PersistenceService db,
```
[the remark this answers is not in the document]
davidhunglam: I'll try to revert these, as well.

srv/src/test/java/my/bookshop/handlers/ReviewServiceHandlerTest.java
```diff
@@ -0,0 +1,40 @@
...
+import com.sap.cds.services.draft.DraftService;
+
+import cds.gen.reviewservice.ReviewService_;
+import cds.gen.reviewservice.Reviews;
+import cds.gen.reviewservice.Reviews_;
+
+@ExtendWith(MockitoExtension.class)
+@ExtendWith(SpringExtension.class)
+@SpringBootTest
+public class ReviewServiceHandlerTest {
```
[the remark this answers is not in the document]
davidhunglam: In AdminServiceTest we have the same, but okay, I can remove it.

app/common.cds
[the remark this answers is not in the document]
beckermarc: Is the Eclipse formatter different from the one in VS Code? I never saw this kind of formatting so far and was wondering about it therefore. It that's like the best practice format induced by the tools we can also keep it like this. For me personally this is taking up too much space, but formatting is a topic where you can have endless discussions..

db/reviews.cds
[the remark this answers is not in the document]
beckermarc: We can also keep the reviewer and date and remove the managed aspect. Also fine with that, if we want to showcase that you can also use @cds.on.insert yourself. Whatever you like better..

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
davidhunglam: Both the Eclipse and the VSCode formatter produce this... But I'll revert it manually.

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
davidhunglam: Will change this.

srv/src/main/java/my/bookshop/handlers/ReviewServiceHandler.java
davidhunglam: It is triggered, yes, even before the enum assertion, it seems. And without it, I was still able to create an invalid rating via the addReview action in the CatalogService.

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
davidhunglam: This is only used in the vue app so far, so I decided to keep it simple (as in the nodejs sample).

srv/review-service.cds
davidhunglam: Yeah, that makes sense for the ReviewService. I'll change that.

srv/src/main/java/my/bookshop/Application.java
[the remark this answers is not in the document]
beckermarc: Can you try if setting `server.servlet.encoding.charset` and `server.servlet.encoding.force` in application.yaml also works? See also https://stackoverflow.com/a/62826669
I fear that adding this thing here so prominently in Application.java class will cause other people to copy and paste this into their apps without thinking about why its required.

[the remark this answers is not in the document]
beckermarc: Interesting issue though. I am wondering if setting the accepted encoding in the request in the Vue.js app might also solve the issue.

app/common.cds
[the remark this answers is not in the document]
davidhunglam: I think the options are the same in both as they're both based on the same language server. But somehow the defaults in Eclipse seem different. I find it easier to set it up correctly in VSCode.

srv/src/main/java/my/bookshop/handlers/ReviewServiceHandler.java
beckermarc: Hm, I guess we need to debug this. This smells like a bug in our enum assertion validation.

srv/src/test/java/my/bookshop/handlers/ReviewServiceHandlerTest.java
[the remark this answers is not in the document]
beckermarc: The intention in AdminService was to simply showcase the usage of `@WithMockUser`. We just thought of a very simple test to do this, which was simply checking that an ID is generated on a new draft. I think we don't need to duplicate this here.

app/common.cds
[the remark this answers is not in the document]
beckermarc: hm, that makes it nearly impractical to use out-of-the-box when working with multiple developers with different IDEs :(
I'm personally no friend of code formatters. In my feeling they never get it quite right, but thats - as said before - a topic of taste and endless discussion :)

db/reviews.cds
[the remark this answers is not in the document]
davidhunglam: I used the managed aspect in the end. It works nicely with the new restrictions in the ReviewService from business perspective. I also changed the Object Page to show modifiedBy and modifiedAt. If the admin now modifies a review of another user, it will show in these fields.

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -62,12 +63,12 @@
 
 	private final Messages messages;
 
-	private final BookRatingService bookRatingService;
+	private final RatingCalculator bookRatingService;
```
beckermarc, line 66: `bookRatingService` -> `reviewCalculator`

srv/src/main/java/my/bookshop/Application.java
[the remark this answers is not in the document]
davidhunglam: Accepted encoding is usually used for compression. There's an `Accepted-Charset` header that I tried to set in Axios that is used for making the requests, but to no effect, unfortunately. Setting the enforcement in the `application.yaml`, however, works.

srv/src/test/java/my/bookshop/handlers/ReviewServiceHandlerTest.java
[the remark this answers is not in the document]
davidhunglam: They're separated now.

srv/src/main/resources/application.yaml
```diff
@@ -1,3 +1,7 @@
+server.servlet.encoding:
```
beckermarc, line 1: I'd move this to the same document in the multi-document YAML, where we also define the static-locations. The setting is only required together with the static-locations config and we could move that in the future to the same profile, if desired. In the cloud deployment the static resources are served by approuter and not by the Spring Boot application anymore.

srv/src/test/java/my/bookshop/handlers/CatalogServiceHandlerTest.java
```diff
@@ -1,25 +1,60 @@
...
 import cds.gen.catalogservice.Books;
+import cds.gen.catalogservice.CatalogService_;
+import cds.gen.catalogservice.Reviews;
+import cds.gen.reviewservice.ReviewService_;
+import my.bookshop.BookRatingService;
 
 @ExtendWith(MockitoExtension.class)
+@ExtendWith(SpringExtension.class)
+@SpringBootTest
 public class CatalogServiceHandlerTest {
```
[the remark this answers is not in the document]
davidhunglam: They are separated now.

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
davidhunglam: Done.

srv/admin-service.cds
```diff
@@ -1,28 +1,45 @@
-using { sap.common.Languages as CommonLanguages } from '@sap/cds/common';
-using { my.bookshop as my } from '../db/schema';
+using
```
beckermarc, line 1: I think this file still uses the old formatting. Can we reformat it with the new settings?

srv/src/test/java/my/bookshop/BookRatingServiceTest.java
```diff
@@ -0,0 +1,37 @@
+package my.bookshop;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.math.BigDecimal;
+import java.util.stream.Stream;
+
+import org.junit.jupiter.api.Test;
+
+public class BookRatingServiceTest {
```
beckermarc, line 10: I'd suggest to rename this to `RatingCalculatorTest`

srv/src/test/java/my/bookshop/CatalogServiceTest.java
```diff
@@ -0,0 +1,152 @@
...
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.security.test.context.support.WithMockUser;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+import cds.gen.catalogservice.AddReviewContext;
+import cds.gen.catalogservice.CatalogService_;
+import cds.gen.catalogservice.Reviews;
+import cds.gen.reviewservice.ReviewService_;
+
+@ExtendWith(MockitoExtension.class)
```
beckermarc, line 28: Mockito is not used in this test..

srv/src/test/java/my/bookshop/CatalogServiceTest.java
```diff
@@ -0,0 +1,152 @@
...
+@ExtendWith(SpringExtension.class)
+@SpringBootTest
+public class CatalogServiceTest {
+
+    @Autowired
```
beckermarc, line 33: Can you please fix the indentation?

srv/src/test/java/my/bookshop/handlers/CatalogServiceHandlerTest.java
```diff
@@ -4,22 +4,38 @@
...
 	private PersistenceService db;
 
+	@Mock
+	private CdsModel model;
+
+	@Mock
+	private Messages messages;
+
+	@Mock
+	private RatingCalculator bookRatingService;
```
beckermarc, line 34: `bookRatingService` -> `ratingCalculator`

srv/cat-service.cds
```diff
@@ -1,10 +1,41 @@
...
+        excluding {
+            descr
+        };
+
+    action submitOrder(book : Books : ID, amount : Integer) returns {
+        stock : Integer
+    };
+
+    // access control restrictions
+    annotate CatalogService.Reviews with @restrict : [
```
beckermarc, line 31: Is the qualifier `CatalogService` required here, as you define this inside the `CatalogService`?

srv/review-service.cds
```diff
@@ -0,0 +1,30 @@
...
+    entity Books   as projection on my.Books excluding {
+        createdBy,
+        modifiedBy
+    }
+
+    @readonly
+    entity Authors as projection on my.Authors;
+
+    // access control restrictions
+    annotate ReviewService.Reviews with @restrict : [
```
beckermarc, line 17: Is the qualifier `ReviewService` required?

beckermarc (commented):
Looks great! There are only a few minor final improvements I think we can make :)

db/reviews.cds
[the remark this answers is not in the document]
beckermarc: Great!

srv/src/main/java/my/bookshop/Application.java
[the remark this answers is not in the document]
beckermarc: Hm, strange that the header doesn't work. But +1 for the config in application.yaml!

srv/src/test/java/my/bookshop/CatalogServiceTest.java
davidhunglam: Weird... locally it does not look like this... I have to make changes here, anyway, and then I'll reformat.

srv/review-service.cds
```diff
@@ -27,4 +27,4 @@ service ReviewService {
     ];
 }
 
-annotate ReviewService.Reviews with @odata.draft.enabled;
+annotate Reviews with @odata.draft.enabled;
```
beckermarc, line 30: I think here the qualifier is still required, as its defined outside the service -> see the test failure

beckermarc: Interesting, that the compiler doesn't throw on this already during compilation..

beckermarc: I think there is a typo: `ReviewsService` -> `ReviewService`, hm, sad that the compiler doesn't check this :(

davidhunglam: Yeah, I just saw it, too.

beckermarc (approved):

PR #69, opened 2021-03-23
Fix annotation name

beckermarc:
Annotation to assert enums is called `@assert.range`, see also https://cap.cloud.sap/docs/guides/generic#assertrange-check-constraints

### DIFF
--- a/db/reviews.cds
+++ b/db/reviews.cds
@@ -19,7 +19,7 @@ entity Reviews : cuid, managed {
 annotate Reviews with {
     subject @mandatory;
     title @mandatory;
-    rating @assert.enum;
+    rating @assert.range;
 }
 
 type Rating : Integer enum {


davidhunglam:
> Annotation to assert enums is called `@assert.range`, see also https://cap.cloud.sap/docs/guides/generic#assertrange-check-constraints

Explains why it didn't work :)

Maybe we should tell the nodejs team about that too, as they do the same mistake in their review sample.

davidhunglam (approved):

beckermarc:
> Maybe we should tell the nodejs team about that too, as they do the same mistake in their review sample.

It looks like they had an `@assert.enum` earlier, which was deprecated in April 2020 (https://cap.cloud.sap/docs/releases/apr20#value-range-annotations-for-odata). Maybe it still works for them. However I guess its a good idea to inform them.

beckermarc:
PR for Node.js sample: https://github.com/SAP-samples/cloud-cap-samples/pull/212

PR #72, opened 2021-03-25
Improved robustness of setIsReviewable

davidhunglam:
(please don't mind the incorrect branch name, it should rather be read-books-handler-robustness)
- we cannot rely on always having the id property of books here
- a request on books(ID=...)/title for example would run into an NPE

### DIFF
--- a/srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
+++ b/srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
@@ -143,7 +143,8 @@ class CatalogServiceHandler implements EventHandler {
 	@After
 	public void setIsReviewable(CdsReadEventContext context, List<Books> books) {
 		String user = context.getUserInfo().getName();
-		List<String> bookIds = books.stream().map(b -> b.getId()).collect(Collectors.toList());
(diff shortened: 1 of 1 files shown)


srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -143,7 +143,8 @@ public void discountBooks(Stream<Books> books) {
 	@After
 	public void setIsReviewable(CdsReadEventContext context, List<Books> books) {
 		String user = context.getUserInfo().getName();
-		List<String> bookIds = books.stream().map(b -> b.getId()).collect(Collectors.toList());
+		List<String> bookIds = books.stream().filter(b -> b.getId() != null).map(b -> b.getId())
+				.collect(Collectors.toList());
```
MarcelMerkle, line 147: Should we add a check if the `List` is empty?

davidhunglam: Yes, I think that would make sense. If `bookIds` is empty we can skip the fetching from the db and setting the field.

beckermarc (approved):

MarcelMerkle (approved):

PR #77, opened 2021-04-12
Use H2 as in-memory database

beckermarc:
To be improved in the future in `cds-services`:
- Auto-detection of H2 as embedded database
- Ability to disable detection of datasources from environment (e.g. `default-env.json`) -> previously covered by explicit `spring.datasource.url` configuration

### DIFF
--- a/.cdsrc.json
+++ b/.cdsrc.json
@@ -17,5 +17,8 @@
     },
     "hana": {
         "deploy-format": "hdbtable"
+    },
+    "sql": {
+        "dialect": "plain"
     }
 }
--- a/README.md
+++ b/README.md
@@ -19,20 +19,19 @@ Welcome to the bookshop-java project. It demonstrates how to build business appl
   - [Using Eclipse](#using-eclipse)
     - [Building and Running](#building-and-running)
   - [Database Setup and Spring Profiles](#database-setup-and-spring-profiles)
-  - [Using a File-Based SQLite Database](#using-a-file-based-sqlite-database)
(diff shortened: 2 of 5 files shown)


agoerler (approved):

beckermarc:
@mofterdinger can you please also have a look at this PR. @agoerler apparently is not a contributer yet 😢 😄

mofterdinger (approved):
As we promote using the H2 database now in the public sample application, we should also consider to change the cds-services-archetype to create initial projects with H2 support.

beckermarc:
Sure, my idea was to pioneer it in the samples first to find any undiscovered issues. If we feel comfortable with H2 we can also adapt our default project generation :)

mofterdinger:
This change has a side effect. If there is a default-env.json laying around from previous tests against a HANA DB, the cap-runtime initialises the datasource with these settings, even I don't choose profile `cloud`. This was not the case in the previous project configuration.

beckermarc:
Yes, this is a known side-effect (see description above). To be discussed if this is desirable or not, see also our internal PR in cds-services on "Improvements for H2". There I added an explicit flag to disable this datasource auto-configuration from default-env.json.

beckermarc:
BTW: These are exactly the reasons for pioneering it in the samples first and not immediately pushing it out with our archetype :)

mofterdinger:
Pioneering is good, but I prefer to do it internally and prepare everything. My comment wasn't about updating the archetype before we gain some experience with replacing Sqlite with H2.

PR #78, opened 2021-04-12
minor: align with https://cap.cloud.sap/docs/java/multitenancy#bindin…

andre68723:
…g-it-together

change srv-url/url -> srv/url to make it consistent to documentation

### DIFF
--- a/mta-multi-tenant.yaml
+++ b/mta-multi-tenant.yaml
@@ -39,9 +39,9 @@ modules:
         properties:
           CDS_MULTITENANCY_APPUI_URL: ~{url}
     provides:
-      - name: srv-url
+      - name: srv
         properties:
-          url: '${default-url}'
+          srv-url: '${default-url}'
 # --------------------- SIDECAR MODULE -----------------------
   - name: bookshop-java-public-sidecar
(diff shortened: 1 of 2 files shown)


beckermarc:
@andre68723 thanks for the contribution and pointing us to that difference!
Looking at that code again I'd personally like it best if it were as simple as `srv/url` in the end. 

@MarcelMerkle any reason why we used `srv` twice in the combined property name? Should we also simplify the documentation or do we want to keep it as is and merge this PR to align the sample with the docs?

beckermarc:
@andre68723 Also mentioned that for consistency we should also adapt `app-url` to simply `app/app-url` or `app/url` depending which way we choose to go.

MarcelMerkle:
@beckermarc No, there is no specific reason. I think it was just like that in the example that I used. We can also use `srv/url` from my perspective.

@andre68723 I think, we should change it in the mta-single-tenant.yaml as well then.

beckermarc (approved):

MarcelMerkle (approved):

beckermarc:
As we know that andre68723 and Andre Voigtmann are the same person I am merging this PR :)

PR #81, opened 2021-04-16
Allow forking a subprocess to support Spring Boot Developer Tools

mofterdinger:
- Disabling forks also disables the Spring Developer Tools in the application
- see: https://docs.spring.io/spring-boot/docs/current/maven-plugin/reference/htmlsingle/#goals-run-parameters-details-fork

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,7 +75,6 @@
 				<artifactId>spring-boot-maven-plugin</artifactId>
 				<version>${spring.boot.version}</version>
 				<configuration>
-					<fork>false</fork>
 					<skip>true</skip>
 				</configuration>
 			</plugin>
--- a/srv/pom.xml
+++ b/srv/pom.xml
@@ -86,6 +86,8 @@
 				<version>${spring.boot.version}</version>
 				<configuration>
 					<skip>false</skip>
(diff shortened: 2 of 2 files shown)


beckermarc:
I remember there was a specific reason for adding this parameter. I think it was related to the working directory and discovery of HTML content via static resources. Can you try that this still works as expected?

mofterdinger:
I tested the application with `mvn spring-boot:run` in project root and I was able to use the fiori application as expected.

beckermarc:
> I tested the application with `mvn spring-boot:run` in project root and I was able to use the fiori application as expected.

I wasn't able to confirm this behaviour. The access to the Fiori UI didn't work, as the working directory was configured to `srv`. I think if we remove fork setting we need to make sure to set the working directory to the project root directory.

mofterdinger:
Ok, strange, I will check it again on my local machine, somehow it worked for me.

mofterdinger:
finally I was able to reproduce and fixed it in srv/pom.xml by configuring a working directory to project root.

beckermarc (approved):
Cool, with this change it works to run mvn spring-boot:run from both the root and the srv directory without any issues. Earlier only the root directory worked correctly with the static Fiori resources.

PR #88, opened 2021-05-12
Annotation path change due to compiler v2 changes

davidhunglam:


beckermarc (changes requested):
I think https://github.com/SAP-samples/cloud-cap-samples-java/blob/c7f9d4ee7dbbb7116881d98f7ab79b6bfb457ab6/app/admin/fiori-service.cds#L90 should be adapted as well..

davidhunglam:
> I think
> 
> https://github.com/SAP-samples/cloud-cap-samples-java/blob/c7f9d4ee7dbbb7116881d98f7ab79b6bfb457ab6/app/admin/fiori-service.cds#L90
> 
> should be adapted as well..

It's not a UI annotation, and the CDS tools give me an error when trying to change this to `.texts`.

beckermarc:
What exactly is the error you are getting? I thought the `@ValueList` annotation is relevant for the UI isn't it?

davidhunglam:
You're right that it's relevant for the UI, but I just meant it's not a `@UI` annotation which is referred to by a reference facet.

In the IDE I got red markers on the locale, text and description fields, saying that they cannot be found. But I just tested running it anyways, and it seems to work, nonetheless.

However, leaving it as `_text` there was no complain by the compiler either... but okay, I'll change it.

app/admin/fiori-service.cds
```diff
@@ -61,9 +61,9 @@ annotate AdminService.Books with @(UI : {
 //
 
 annotate my.bookshop.Books with @fiori.draft.enabled;
-annotate AdminService.Books with @odata.draft.enabled;
+annotate AdminService.Books with @odata.draft.enabled;
```
beckermarc, line 64:
```suggestion
annotate AdminService.Books with @odata.draft.enabled;
```

beckermarc (approved):

PR #89, opened 2021-05-20
Added book import functionality (media data streaming example)

davidhunglam:
- added CSV singleton entity
- implemented singleton initialization on application start-up
- added event handler parsing CSV files to add books to the bookshop
- added more authors to the pre-seeded data
- added sample `CSV` file to use with the book import functionality
- added custom UI extension for the book import button on the `Manage Books` view
- added info to readme


pom.xml
```diff
@@ -16,7 +16,7 @@
 
 		<!-- DEPENDENCIES VERSION -->
 		<jdk.version>1.8</jdk.version>
-		<cds.services.version>1.15.3</cds.services.version>
+		<cds.services.version>1.16.0</cds.services.version>
 		<spring.boot.version>2.4.4</spring.boot.version>
```
mofterdinger, line 20: please also bump version of spring-boot to 2.4.5

davidhunglam: ok, will do

srv/src/main/resources/messages.properties
```diff
@@ -9,4 +9,5 @@ bupa.missing = No Business Parter for this user available
 review.invalid.rating = Rating needs to be within 1 and 5
 review.added = Review added
 order.exceeds.stock = {0} exceeds stock for book
-review.add.forbidden=User not allowed to add more than one review for a given book
\ No newline at end of file
+review.add.forbidden=User not allowed to add more than one review for a given book
+book.import.failed = Import of books failed
```
beckermarc, line 13: Can you please also add the german translation for this to the messages_de?

srv/src/main/java/my/bookshop/handlers/SingletonInitialization.java
```diff
@@ -0,0 +1,29 @@
...
+public class SingletonInitialization implements EventHandler {
+
+	@Autowired
+	private PersistenceService db;
+	
+	@After(event = ApplicationLifecycleService.EVENT_APPLICATION_PREPARED)
+	public void initializeCsvEntity() {
+		Csv csv = Csv.create();
+		csv.setDataType("text/csv");
+		db.run(Upsert.into(Bookshop_.CSV).entry(csv));
```
beckermarc, line 27: This approach is a little problematic with a multi-tenant application. Most likely it will also break the multi-tenant scenario, as in that case it is not clear in which database container to insert the data.

What about doing that via CSV initialization files as well? They get deployed on every tenant database during tenant onboarding.

Alternatively we could also try to implement this fully without a database persistence (which is not really required). Most likely we just need to return a static Csv entity for all READ events on this entity on the Admin Service.

srv/src/main/java/my/bookshop/handlers/AdminServiceHandler.java
```diff
@@ -248,6 +256,39 @@ public void addBookToOrder(AddToOrderContext context) {
 		context.setResult(updatedOrder);
 	}
 
+	@On(event = CdsService.EVENT_UPDATE)
+	public void addBooksViaCsv(CdsUpdateEventContext context, List<Csv> csvs) {
```
beckermarc, line 260: As it is a singleton we could also argue that we always expect only a single Csv in the event handler.

srv/src/main/java/my/bookshop/handlers/AdminServiceHandler.java
```diff
@@ -248,6 +256,39 @@ public void addBookToOrder(AddToOrderContext context) {
...
+						// separate transaction per line
+						context.getCdsRuntime().changeSetContext().run(ctx -> {
+							db.run(Upsert.into(Bookshop_.BOOKS).entry(book));
+						});
+					});
+				} catch (IOException e) {
+					messages.error(MessageKeys.BOOK_IMPORT_FAILED);
+				}
+			} else {
+				db.run(Upsert.into(Bookshop_.CSV).entry(csv));
```
beckermarc, line 286: We could also just ignore this, especially if we want to get rid of the unnecessary db persistence for this entity.

db/books.cds
```diff
@@ -8,6 +8,11 @@ using {
     cuid
 } from '@sap/cds/common';
 
+entity Csv @odata.singleton {
```
beckermarc, line 11: As already suggested earlier: We don't really want a database persistence for this. Conceptually this better fits as an entity that is directly defined in the admin service and is annotated with @cds.persistence.skip to prevent any database table to be created for it.

beckermarc (changes requested):
Nice change! I suggest we try to get rid of the database persistence for the Csv entity, as this is not required and currently rather acts as a workaround.
This would also make a nice showcase for non db-based entities in our sample!

db/books.cds
davidhunglam: That's a great option that I was not aware of. Thank you for the suggestion.

srv/admin-service.cds
```diff
@@ -9,6 +9,12 @@ service AdminService @(requires : 'admin') {
 
   entity Authors as projection on my.Authors;
   entity Orders  as select from my.Orders;
+
+  @cds.persistence.skip
+  entity Csv @odata.singleton {
+    data     : LargeBinary @Core.MediaType : dataType;
+    dataType : String @Core.IsMediaType;
```
beckermarc, line 16: Is there still a reason for this element, as it's value is anyway always expected to be `text/csv`? 
Could we also use `@Core.MediaType: 'text/csv'` on data directly?

srv/src/main/java/my/bookshop/handlers/AdminServiceHandler.java
```diff
@@ -248,6 +260,43 @@ public void addBookToOrder(AddToOrderContext context) {
 		context.setResult(updatedOrder);
 	}
 
+	@On(entity = Csv_.CDS_NAME)
+	public void getCsvSingleton(CdsReadEventContext context) {
+		Csv csv = Csv.create();
+		csv.setDataType("text/csv");
+		Result result = ResultBuilder.selectedRows(Arrays.asList(csv)).result();
```
beckermarc, line 267: you can also directly return the Result object, with this you don't need to use `context.getResult()`. If you remove the context from the arguments you need to add the READ event to the `@On` annotation though :)

srv/admin-service.cds
davidhunglam: Yeah, we could. That would make sense, I'll change it.

srv/src/main/java/my/bookshop/handlers/AdminServiceHandler.java
davidhunglam: The `Result` object? Or can I also directly return the `Csv` object? I just tried

```java
@On(entity = Csv_.CDS_NAME, event = CdsService.EVENT_READ)
public Csv getCsvSingleton() {
    return Csv.create();
}
```
and that seems to work, too.

beckermarc: Both is fine :) That way its even more elegant ;)

beckermarc (approved):

PR #92, opened 2021-06-08
Minor improvements

davidhunglam:
- error handling
- naming


app/admin/webapp/extension/UploadDialog.fragment.xml
```diff
@@ -7,7 +7,7 @@
                     id="uploader"
                     fileType="csv"
                     multiple="false"
-                    uploadUrl="/api/admin/Csv/data"
+                    uploadUrl="/api/admin/CsvStream/data"
```
beckermarc, line 10: From a URL perspective I think having something like `/api/admin/upload/csv` would be nice, but I guess any naming here would be fine for me.

srv/src/main/resources/messages_de.properties
```diff
@@ -10,4 +10,5 @@ review.invalid.rating = Die Bewertung muss ein Wert innerhalb 1 und 5 sein
 review.added = Bewertung hinzugef\u00FCgt
 order.exceeds.stock = {0} ist mehr als f\u00FCr das Buch auf Vorrat \u00FCbrig ist
 review.add.forbidden=Es ist nicht mehr als eine Bewertung pro Buch erlaubt
-book.import.failed = B\u00FCcher Import fehlgeschlagen
\ No newline at end of file
+book.import.failed = B\u00FCcher Import fehlgeschlagen
+book.import.invalid.csv = Die CSV Datei enthält eine ungültige Struktur - Bitte \u00FCberpr\u00FCfen Sie den Inhalt
```
beckermarc, line 14: Do we still have issues with umlauts? The fact that you only replaced half of the umlauts maybe suggests that it is not required anymore to do this?

davidhunglam: No, I just forgot to change it when I tried out different phrases. I have not tested with German locale, I just assumed it works as suggested by the existing entries.

app/admin/webapp/extension/UploadDialog.fragment.xml
davidhunglam: Upload sounds good, too, as we are not persisting the stream anyways. So there'd be no "download". But entities in the model have to start with a capital, right? Is there a way to override the entity name for the path?

srv/src/main/resources/messages_de.properties
beckermarc: Okay :) I was always wondering if this is still necessary and just tested it and it is not. I think we fixed that bug long ago, that we couldn't handle umlauts there. I will open a separate PR. You can use normal umlauts for your new error texts.

app/admin/webapp/extension/UploadDialog.fragment.xml
beckermarc: > But entities in the model have to start with a capital, right?

I don't think anybody enforces this, but I am not sure.

> Is there a way to override the entity name for the path only?

No currently not.

srv/src/main/resources/messages_de.properties
davidhunglam: okay :)

app/admin/webapp/extension/UploadDialog.fragment.xml
davidhunglam: > I don't think anybody enforces this, but I am not sure.

Actually, it works. The generated class will be annotated like this:
`@CdsName("AdminService.upload")`

But do we want to promote this?

srv/src/main/resources/messages_de.properties
beckermarc: https://github.com/SAP-samples/cloud-cap-samples-java/pull/93

app/admin/webapp/extension/UploadDialog.fragment.xml
beckermarc: > But do we want to promote this?

good question, not sure about that :)

davidhunglam: Playing it safe then :)

srv/src/main/java/my/bookshop/handlers/AdminServiceHandler.java
```diff
@@ -261,9 +261,9 @@ public void addBookToOrder(AddToOrderContext context) {
 	/**
 	 * @return the static CSV singleton upload entity
 	 */
-	@On(entity = Csv_.CDS_NAME, event = CdsService.EVENT_READ)
-	public Csv getCsvSingleton() {
-		return Csv.create();
+	@On(entity = Upload_.CDS_NAME, event = CdsService.EVENT_READ)
+	public Upload getCsvSingleton() {
```
beckermarc, line 265:
```suggestion
	public Upload getUploadSingleton() {
```

beckermarc (approved):
LGTM

AndreasSchoesser (approved):

PR #107, opened 2021-08-13
Improve Addresses S/4 integration

beckermarc:
- Adopted new best practices in CDS model external.cds
- Added new mocked profile, which uses a Remote Service configured to a mocked API_BUSINESS_PARTNER Application Service
- Ensure mock tables are never present when running with the cloud profile
- Only use AdminService.Addresses projection in event handler, projections are automatically resolved by downstream services
- Code robustness improvements
- Added tests for Shipping Addresses integration scenario


srv/src/main/java/my/bookshop/handlers/DestinationConfiguration.java
```diff
@@ -18,15 +22,32 @@
...
+		Integer port = environment.getProperty("local.server.port", Integer.class);
+		if(port != null) {
+			DefaultHttpDestination httpDestination = DefaultHttpDestination
+			.builder("http://localhost:" + port)
+			.name("myself").build();
+
+			DestinationAccessor.prependDestinationLoader(
+				new DefaultDestinationLoader().registerDestination(httpDestination));
 		}
 	}
```
davidhunglam, line 52: For the `cloud` profile, is this guaranteed to happen before `initializeDestinations()`?

srv/external.cds
```diff
@@ -30,12 +17,20 @@ extend service external with {
...
-entity my.bookshop.Addresses as SELECT from external.Addresses { *, false as tombstone: Boolean };
+entity my.bookshop.Addresses as projection on external.A_BusinessPartnerAddress {
+  key AddressID as ID,
+  key BusinessPartner as businessPartner,
+  @readonly Country as country,
+  @readonly CityName as city,
+  @readonly PostalCode as postalCode,
+  @readonly StreetName as street,
+  @readonly HouseNumber as houseNumber,
+  false as tombstone: Boolean
```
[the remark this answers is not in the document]
beckermarc: unfortunately not really, the default syntax is not allowed in projections. I'd also prefer that, but its currently not possible.
Currently this only adds the field to the table, but setting the value is currently done in custom code.

srv/src/main/java/my/bookshop/handlers/DestinationConfiguration.java
beckermarc: this most likely happens after `initializeDestinations()` do you see an issue with that?

davidhunglam: Okay, first I thought they might overwrite each other, but they are actually just added as independent loaders to the DestinationAccessor.

But why did you not just add the new destination loader within the before handler, instead of creating a new event listener for the `ApplicationReadyEvent` event?

davidhunglam: Or is the `port` property eventually not available yet at that time?

beckermarc: > Or is the local.server.port property eventually not available yet at that time?

Exactly :) the Tomcat port is only available very late in the game :)

davidhunglam (approved):

PR #124, opened 2021-11-25
Fix post path in Vue app; update dependencies in pom.xml

ruzickam:
Note:
When a review is added or updated in the Vue app (http://localhost:8080/vue/index.html), the method `afterAddReview(...)` in `ReviewsServiceHandler` is not triggered. I cannot figure out why.

### DIFF
--- a/app/vue/app.js
+++ b/app/vue/app.js
@@ -2,7 +2,7 @@
 const $ = (sel) => document.querySelector(sel);
 const GET = (url) => axios.get("/api/ReviewsService" + url);
 const PUT = (cmd, data) => axios.patch("/api/ReviewsService" + cmd, data);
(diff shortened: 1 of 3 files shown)


beckermarc:
@dimamost can you please also check @ruzickam's comment about `ReviewsServiceHandler.afterAddReview` not being called? Looks to me as if the event handler is registered on the wrong entity..

@ruzickam thanks for the contribution 🚀

ruzickam:
@beckermarc I found the reason why it didn't work. There were wrong inputs for `Reviews` and `Reviews_` interfaces. New commit added.

beckermarc (approved):
LGTM! Thanks for the contribution!

PR #126, opened 2021-12-07
Update CDS Services to 1.21.0

beckermarc:


davidhunglam (approved):

pom.xml
```diff
@@ -66,6 +67,33 @@
 				<artifactId>cf-java-logging-support-logback</artifactId>
 				<version>${cf-java-logging-support.version}</version>
 			</dependency>
+
+			<!-- XSUAA -->
+			<dependency>
```
BraunMatthias, line 72: Should we deomnstrate the new XSUAA bom pom here?

beckermarc: good idea, I will create a PR..

PR #143, opened 2022-06-02
Explicit checks are replaced with annotations

vmikhailenko:
Explicit checks on managed to-one associations are replaced by a annotations `@mandatory` and `@assert.target`.
The ReviewServiceHandler class removed and replaced with them completely.


srv/src/main/java/my/bookshop/handlers/ReviewServiceHandler.java
```diff
@@ -1,54 +0,0 @@
...
-		reviews.forEach(review -> {
-			validateBook(review);
-			validateRating(review);
-		});
-	}
-
-	private void validateRating(Reviews review) {
-		Integer rating = review.getRating();
-		if (rating == null || rating < 1 || rating > 5) {
-			messages.error(MessageKeys.REVIEW_INVALID_RATING).target("in", Reviews_.class, r -> r.rating());
```
beckermarc, line 44: I guess we can also remove now various unused constants from `MessageKeys` and resource bundles now.

db/reviews.cds
```diff
@@ -9,7 +9,7 @@ using my.bookshop.Books from './books';
 
 entity Reviews : cuid, managed {
     @cds.odata.ValueList
-    book     : Association to Books;
+    book     : Association to Books @mandatory @assert.target;
```
beckermarc, line 12: Maybe move this down to line 19++ where also the other input validation annotations can be found

srv/src/test/java/my/bookshop/AdminServiceTest.java
```diff
@@ -31,6 +34,9 @@ public class AdminServiceTest {
 	@Resource(name = AdminService_.CDS_NAME)
 	private DraftService adminService;
 
+	@Resource(name = AdminService_.CDS_NAME)
+	private ApplicationService activeService;
```
beckermarc, line 38: Why not simply use the existing `adminService` instance?

beckermarc (changes requested):

srv/src/main/java/my/bookshop/handlers/AdminServiceHandler.java
```diff
@@ -122,24 +117,25 @@ public void beforeCreateOrder(Stream<Orders> orders, EventContext context) {
 
 					// check if enough books are available
 					Result result = db.run(Select.from(BOOKS).columns(b -> b.ID(), b -> b.stock(), b -> b.price()).byId(bookId));
-					Books book = result.first(Books.class).orElseThrow(notFound(MessageKeys.BOOK_MISSING));
```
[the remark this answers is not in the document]
vmikhailenko: It is replaced with a combination of annotations. It could remain, but should never fire.

srv/src/main/java/my/bookshop/handlers/ReviewServiceHandler.java
vmikhailenko: Done. BOOK_MISSING still used in action.

db/reviews.cds
vmikhailenko: Done.

srv/src/test/java/my/bookshop/AdminServiceTest.java
vmikhailenko: I wanted to bypass the drafts. Changed to DraftService.run().

beckermarc:
Should we also add the configuration for the actual db-level constraint checks generated by the CDS Compiler? I think we shouldn't promote `@assert.target` without that. `@assert.target` doesn't really safely guarantee that the reference actually exists. Parallel transactions can still create inconsistent data. It mainly helps to improve the error message in case of errors.

vmikhailenko:
> Should we also add the configuration for the actual db-level constraint checks generated by the CDS Compiler? I think we shouldn't promote `@assert.target` without that. `@assert.target` doesn't really safely guarantee that the reference actually exists. Parallel transactions can still create inconsistent data. It mainly helps to improve the error message in case of errors.

These ones https://cap.cloud.sap/docs/guides/providing-services#refs?

vmikhailenko:
> No, I meant this one: [cap.cloud.sap/docs/guides/databases#db-constraints](https://cap.cloud.sap/docs/guides/databases#db-constraints)

As far as I understood, the intention of assert.target was to provide limited replacement for these in case they are disabled or have them as a last line of defence if they enabled.

beckermarc:
I think the feature is most valuable when database constraints are generally enabled, but certain error messages should be improved. The @assert.target can not guarantee that the data never gets inconsistent. This is only guaranteed with database constraints.

vmikhailenko:
> I think the feature is most valuable when database constraints are generally enabled, but certain error messages should be improved. The @assert.target can not guarantee that the data never gets inconsistent. This is only guaranteed with database constraints. If it is not a big task we could enable the database constraints as well. If it turns out more complicated we can tackle that in a separate change.

Added the cds.assert.integrity.

srv/src/main/resources/application.yaml
```diff
@@ -18,6 +18,8 @@ cds:
         - "10401010"
     - name: user
       password: user
+  features:
+    assert_integrity: db
```
beckermarc, line 22: This is not a valid configuration in application.yaml.

beckermarc: Let's skip the actual assertions on db level for now. I think it might be valuable to showcase this here in the bookshop, but I guess this is a bigger change..

vmikhailenko: cdsrc.json?

beckermarc (approved):

PR #144, opened 2022-06-10
Enable access to localhost:8080/h2

gicalin:
Small technical contribution to be able to see and change entries directly in h2 while the apps are running.


beckermarc (commented):
Thanks for the contribution!! :)

I am however not sure, if we want to promote adding the h2 console path to the public endpoints by default. Another thing to consider: We have improved the support for Spring Boot Dev Tools with our latest CAP Java release. When adding Spring Boot Dev Tools the H2 Console is automatically enabled during development. Therefore these changes are no longer necessary when we add this dependency (https://github.com/SAP-samples/cloud-cap-samples-java/pull/145)

The security config concern could however also be solved by introducing a dedicated security config that is only available in the default profile.

gicalin:
Closing since I understand the change is not needed.

PR #150, opened 2022-06-30
Add single page index pages

uwe-klinger:
This allows to test the UIs quickly with the central approuter (e.g. using Launchpad Service) without the need of any UI configuration.

### DIFF
--- a/app/addresses/webapp/index.html
+++ b/app/addresses/webapp/index.html
@@ -12,7 +12,7 @@
     </style>
     <script
         id="sap-ui-bootstrap"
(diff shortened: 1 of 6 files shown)


beckermarc:
In the `fiori.html` we decided to not fix the UI5 version. I think we should use the latest one then also in these files.

uwe-klinger:
> In the `fiori.html` we decided to not fix the UI5 version. I think we should use the latest one then also in these files.

Ok, I've updated the PR. Please recheck.

beckermarc (approved):

PR #152, opened 2022-06-30
Helm Chart

uwe-klinger:


srv/pom.xml
```diff
@@ -104,6 +120,19 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.sap.cloud.security.xsuaa</groupId>
+            <artifactId>xsuaa-spring-boot-starter</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.sap.cds</groupId>
+            <artifactId>cds-feature-hana</artifactId>
+        </dependency>
```
beckermarc, line 135: Those dependencies automatically come with cds-starter-cloudfoundry. No need to add them here..

srv/src/main/java/my/bookshop/config/DestinationConfiguration.java
```diff
@@ -20,13 +21,16 @@ public class DestinationConfiguration {
...
+		Boolean onPremise = environment.getProperty("on.premise", Boolean.class);
 		if(port != null && destinationName != null) {
 			DefaultHttpDestination httpDestination = DefaultHttpDestination
 			.builder("http://localhost:" + port)
 			.name(destinationName).build();
 
 			DestinationAccessor.prependDestinationLoader(
 				new DefaultDestinationLoader().registerDestination(httpDestination));
+		} else if (destinationName == "s4-destination" && onPremise) {
+			DestinationAccessor.getDestination("s4-destination").asHttp().decorate(DefaultErpHttpDestination::new);
```
beckermarc, line 33: I think this doesn't work. There is currently a workaround required for on-premise destinations, which is described here: https://github.wdf.sap.corp/MA/cap-cloud-sdk-collaboration/issues/55#issuecomment-2948646

srv/pom.xml
uwe-klinger: Ok, I'll remove them and retry.

srv/src/main/java/my/bookshop/config/DestinationConfiguration.java
uwe-klinger: As discused, the code doesn't work.  The header field must be configured in the destination.

srv/pom.xml
```diff
@@ -61,6 +67,16 @@
 			<artifactId>scp-cf</artifactId>
 		</dependency>
 
+		<!-- CLOUD SDK for on-premise connectivity -->
```
uwe-klinger, line 70: @beckermarc: I wonder if we still need it. I could retry without.

beckermarc: I guess we do, if you want to connect to on premise systems. But you can try out what happens without these dependencies. AFAIK they are however responsible for implementing the connectivity service stuff, which is not required for cloud S/4 hana systems. Not sure if the setup to connect to an on-premise S/4 needs to be part of the sample though.

beckermarc (approved):

PR #155, opened 2022-07-06
Update to cds-services 1.26.0

BraunMatthias:


beckermarc (commented):
Tests might be failing, because of default authenticated access now?

BraunMatthias:
yes, it seems I've to adapt the tests as well

PR #159, opened 2022-07-29
Add links to Kyma tool install pages

beckermarc:


uwe-klinger (approved):
Alternatively, we could point to [our tutorial](https://developers.sap.com/tutorials/btp-app-kyma-prepare-dev-environment.html#e5f7eb1e-fd40-41c3-b67d-268fe8a41474). We could also consider to have a specific tutorial for just the preconditions for deploying an app. That would only require a bit of refactoring.

beckermarc:
The tutorial is already linked in the tip above the list of tools :)

davidhunglam (approved):

PR #166, opened 2022-09-02
Update of build configuration for MTX sidecar

vmikhailenko:
The configuration of build for MTX sidecar updated per recommendations for cds-dk 6.1+

### DIFF
--- a/mta-multi-tenant.yaml
+++ b/mta-multi-tenant.yaml
@@ -4,12 +4,6 @@ version: 1.0.0
 description: "Multitenant Bookshop CAP Java Project with UI"
 parameters:
   enable-parallel-deployments: true
-build-parameters:
-  before-all:
-   - builder: custom
-     commands:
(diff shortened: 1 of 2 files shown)


mta-multi-tenant.yaml
```diff
@@ -48,10 +48,15 @@ modules:
...
-    path: mtx/sidecar/gen
+    path: mtx/sidecar
     parameters:
       memory: 256M
       disk-quota: 512M
+    build-parameters:
+      builder: custom
+      build-result: gen
+      commands:
+        - npm run build
```
dimamost, line 59: As we have this build step, do we need the pre-build step in the line 11 (- npm install --prefix mtx/sidecar --production)? This would be really good if we could get rid of it, as it works only properly on Mac.

vmikhailenko: Good catch! It seems that it is not needed anymore.

PR #171, opened 2022-10-21
use dialect plain to fix h2 issue

mofterdinger:


mofterdinger:
For some reason it's required to use sql dialect `plain` again to get the bookshop running with h2.

agoerler (approved):

PR #173, opened 2022-10-31
Update README

PalaniSockalingam:
Updated the build part in README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,18 +239,6 @@ Before you can access the UI using the (tenant-specific) URL to the bookshop(-mt
 - Logged into Kyma Runtime (with `kubectl` CLI), Cloud Foundry space (with `cf` CLI) and Container Registry (with `docker login`)
 - `@sap/cds-dk` >= 6.0.1
 
-### Add Deployment Files
-
-CAP tooling provides your a Helm chart for deployment to Kyma.
-
-Add the CAP Helm chart with the required features to this project:
-
-```bash
-cds add helm:hana_deployer
-cds add helm:xsuaa
-cds add helm:html5_apps_deployer
-```
-
 #### Helm chart configuration
 
 This project contains a pre-configured configuration file `values.yaml`, you just need to do the following changes in this file:
@@ -327,7 +315,8 @@ cds build --production
 pack build $YOUR_CONTAINER_REGISTRY/bookshop-hana-deployer \
      --path db \
      --buildpack gcr.io/paketo-buildpacks/nodejs \
-     --builder paketobuildpacks/builder:base
+     --builder paketobuildpacks/builder:base \
+     --env BP_NODE_RUN_SCRIPTS= ""
 ```
 
 (Replace `$YOUR_CONTAINER_REGISTRY` with the full-qualified hostname of your container registry)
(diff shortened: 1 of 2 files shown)


beckermarc:
Hi @PalaniSockalingam, thanks for opening this PR!
Can you please explain the reasons behind the changes you are suggesting so that we can better understand the change?
Did you run into issues with the current ordering of steps?
Did you have issues with the `values.yaml` file, if yes which?

README.md
```diff
@@ -239,18 +239,69 @@ Before you can access the UI using the (tenant-specific) URL to the bookshop(-mt
...
+**Build data base deployer image:**
+
+```
+cds build --production
+
+pack build $YOUR_CONTAINER_REGISTRY/bookshop-hana-deployer \
+     --path db \
+     --buildpack gcr.io/paketo-buildpacks/nodejs \
+     --builder paketobuildpacks/builder:base \
+     --env BP_NODE_RUN_SCRIPTS= ""
```
phil9909, line 273:
```suggestion
     --env BP_NODE_RUN_SCRIPTS=
```

README.md
```diff
@@ -239,18 +239,69 @@ Before you can access the UI using the (tenant-specific) URL to the bookshop(-mt
...
 
 Add the CAP Helm chart with the required features to this project:
 
 ```bash
-cds add helm:hana_deployer
-cds add helm:xsuaa
-cds add helm:html5_apps_deployer
+cds add helm
+cds add xsuaa
+cds add html5-5-repo
```
sidakphull, line 302: this should be `cds add html5-repo`

sidakphull:
> Can you please explain the reasons behind the changes you are suggesting so that we can better understand the change?
> Did you run into issues with the current ordering of steps?

Hi @beckermarc,

The reason we moved the build steps before the `cds add ...` commands is that sometimes `cds add ..` command will make a change to the module which can cause issues while building the image.

For example: `cds add hana` adds a build script to the `package.json` present inside db folder. This started to cause errors while building `hana-deployer` image which we were able to fix by adding the `--env BP_NODE_RUN_SCRIPTS=` flag.

Even though we found a fix for this particular error we just wanted to move the build images step before the `cds add ..` commands so that we won't face these kind of issues in the future. (We have made this change for our Kyma E2E tests as well.)

HTML5 Application deployer image is still built afterwards because it requires information from the `values.yaml`. Soon we will be replacing the HTML5 deployment part for this sample with Approuter Deployer and image for that will be built before the `cds add ..` commands only.

Now we only face this issue in this application because this application is **missing `cds.requires` section inside `.cdsrc.json`**. So when we call `cds add helm` it isn't able to determine what all features(hana, approuter etc.) this application requires and that is why we have to call `cds add hana`, `cds add xsuaa` again to add the configuration required for deploying/binding these on Kyma. This isn't `cds add helm` specific. If you call `cds add mta` it also won't be able to determine the resources required due to the same reason.

Thanks
Sidak

beckermarc:
> The reason we moved the build steps before the cds add ... commands is that sometimes cds add .. command will make a change to the module which can cause issues while building the image.

I think that this needs to be avoided overall. I'd expect that after making changes to the application the build is later executed again and again. If it only works the first time, because it got corrupted by the cds add commands this will not be possible. That means it is not possible to deliver any updates to the application. I think the root cause of cds add corrupting the projects need to be fixed.

> Now we only face this issue in this application because this application is missing cds.requires section inside .cdsrc.json.

This is expected. CAP Java applications don't require a `cds.requires` section, as this section is not evaluated in CAP Java application. The relevant counterparts of this are stored in `application.yaml` in CAP Java applications. I think `cds` commands that want to be compatible with Java should not rely on `cds.requires` being present.

PR #175, opened 2022-11-03
Update CDS Services to 1.29.0

beckermarc:
- Use H2 v2 with new h2 dialect provided by CDS Compiler
- Use new name java instead of old java-cf for build tasks
- Update dependencies

### DIFF
--- a/.cdsrc.json
+++ b/.cdsrc.json
@@ -3,7 +3,7 @@
         "target": ".",
         "tasks": [
             {
-                "for": "java-cf"
+                "for": "java"
             },
             {
                 "for": "mtx",
@@ -17,8 +17,5 @@
     },
     "hana": {
         "deploy-format": "hdbtable"
-    },
-    "sql": {
-        "dialect": "plain"
     }
 }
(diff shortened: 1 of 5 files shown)


.cdsrc.json
```diff
@@ -3,7 +3,7 @@
         "target": ".",
         "tasks": [
             {
-                "for": "java-cf"
+                "for": "java"
```
BraunMatthias, line 6: are you sure about this? We still seem to run classic mtx in the bookshop

beckermarc: I think this change is independent of streamlined or classic MTX, but we can also keep it back if you like.

beckermarc: I'd suggest to go for it. @mofterdinger also proposed a similar change to the projects that we generate in the cds-services archetype.

BraunMatthias: indeed` java-cf` and `java `point to the same implementation. `java-cf` should be replaced.

BraunMatthias (approved):

PR #179, opened 2022-11-15
Add mtdeployment to authorities

beckermarc:
As documented here: https://cap.cloud.sap/docs/java/multitenancy#xsuaa-mt-configuration


BraunMatthias (approved):
we should think about switching to mtxs in the default branch soon

PR #182, opened 2022-11-30
Streamlined mtx merge main

BraunMatthias:


beckermarc:
If you do this via a PR and squash merge the merge commit the history will be broken. I think I would just push this change directly to the `streamlined-mtx` branch without opening a PR at all..

PR #183, opened 2022-12-06
Upated mta.yamls

mofterdinger:
align mta-single-tenant.yaml with command: cds add mta

### DIFF
--- a/db/package.json
+++ b/db/package.json
@@ -3,7 +3,11 @@
   "dependencies": {
     "@sap/hdi-deploy": "^4"
   },
+  "engines": {
+    "node": "^16"
+  },
   "scripts": {
-    "start": "node node_modules/@sap/hdi-deploy --exit"
+    "start": "node node_modules/@sap/hdi-deploy/deploy.js",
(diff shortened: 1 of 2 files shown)


BraunMatthias (approved):

db/package.json
```diff
@@ -3,7 +3,11 @@
   "dependencies": {
     "@sap/hdi-deploy": "^4"
   },
+  "engines": {
+    "node": "^16"
+  },
   "scripts": {
-    "start": "node node_modules/@sap/hdi-deploy --exit"
+    "start": "node node_modules/@sap/hdi-deploy/deploy.js",
+    "build": "npm i && npx -p @sap/cds-dk@^6 cds build .. --for hana --production"
```
beckermarc, line 11: Little unfortunate that we hide another @sap/cds-dk version here again. This makes it not really transparent in ensuring that Java app and DB module are built with the same @sap/cds-dk version..

mofterdinger: Yes, you are right, but that's the current situation. Maybe we have to remove the approach to set the cds-dk version in pom.xml and maintain it globally in a package.json that we can share the same cds-dk version for all modules (db, srv, ....)

PR #184, opened 2022-12-09
remove cds-dk installation from build script of db

mofterdinger:
This PR removes the @sap/cds-dk installation from the build script in db/package.json. Now, the build script relies on a previous local installation of the cds-dk, e.g. done with a maven build. Or a globally installed cds-dk version is used.
The mbt build works fine, because first the srv module is built, then the db module.

### DIFF
--- a/db/package.json
+++ b/db/package.json
@@ -8,6 +8,6 @@
   },
   "scripts": {
     "start": "node node_modules/@sap/hdi-deploy/deploy.js",
-    "build": "npm i && npx -p @sap/cds-dk@^6 cds build .. --for hana --production"
+    "build": "npm i && npx cds build .. --for hana --production"
   }
 }
--- a/mta-single-tenant.yaml
+++ b/mta-single-tenant.yaml
@@ -40,6 +40,8 @@ modules:
       builder: custom
       commands:
         - npm run build
+      requires:
+        - name: bookshop-srv
     requires:
       - name: bookshop-hdi-container
 # --------------------- APPROUTER MODULE ---------------------


beckermarc:
> The mbt build works fine, because first the srv module is built, then the db module.

Is this order modelled somewhere explicitly?

mofterdinger:
> > The mbt build works fine, because first the srv module is built, then the db module.
> 
> Is this order modelled somewhere explicitly?

I'm not sure, but it looks like it the order of the modules in the mta.yaml: srv, db, app. At least it's my experience with this project and mbt.

mofterdinger:
Verified it, it's the order of the modules in the mta. I change the order to db, srv, app. Now, first the db module is built. Since the node_modules folder is not existing, the globally installed cds-dk is used:
```
[cds] - building project [/Users/d032842/SAPDevelop/git/cloud-cap-samples-java], clean [true]
[cds] - cds [6.3.2], compiler [3.4.4], home [/usr/local/lib/node_modules/@sap/cds-dk/node_modules/@sap/cds]

```

beckermarc:
Can this be modelled somehow explicitly? Not sure, if this is considered an implementation detail by mbt build or if this is documented behaviour. We need to be careful to rely on that if this behaviour is not considered as "API"

mofterdinger:
> Can this be modelled somehow explicitly? Not sure, if this is considered an implementation detail by mbt build or if this is documented behaviour. We need to be careful to rely on that if this behaviour is not considered as "API"

Yes, it's possible to maintain a build order of the modules. Further details can be found here: https://sap.github.io/cloud-mta-build-tool/configuration/#defining-module-build-order

beckermarc (approved):
I'm fine with this, not sure what @BraunMatthias thinks, as we now deviate again from the defaults :D

PR #189, opened 2022-12-20
Make streamlined MTX new default

beckermarc:


beckermarc:
Tested:
- Single Tenant locally successfully
- Multi Tenant locally successfully
- Single Tenant in BTP successfully
- Multi Tenant in Cloud failed -> cds.xt.Extensions could not be found, seems not to be deployed

README.md
```diff
@@ -45,6 +45,8 @@ Although CAP provides generic **event handlers** to serve most CRUD requests out
 
 A SAP Fiori UI is added using predefined SAP Fiori elements templates. **[SAP Fiori annotations](https://cap.cloud.sap/docs/advanced/fiori#fiori-annotations)** add information to the service definitions, on how to render the data.
 
+CAP provides built-in multitenancy support with out-of-the box tenant isolation. The sample application demonstrate usage of MTX sidecar based on [streamlined MTX](https://cap.cloud.sap/docs/guides/deployment/as-saas?impl-variant=java) and can be deployed as multitenant application. The [deprecated classic MTX](https://cap.cloud.sap/docs/java/multitenancy) setup is shown in the [mtx-classic branch](https://github.com/SAP-samples/cloud-cap-samples-java/tree/mtx-classic) for reference.
```
BraunMatthias, line 48:
```suggestion
CAP provides built-in multitenancy support with out-of-the box tenant isolation. The sample application demonstrates usage of MTX sidecar based on [streamlined MTX](https://cap.cloud.sap/docs/guides/deployment/as-saas?impl-variant=java) and can be deployed as multitenant application. The [deprecated classic MTX](https://cap.cloud.sap/docs/java/multitenancy) setup is shown in the [mtx-classic branch](https://github.com/SAP-samples/cloud-cap-samples-java/tree/mtx-classic) for reference.
```

BraunMatthias (approved):

srv/src/main/java/my/bookshop/handlers/AdminServiceAuditHandler.java
```diff
@@ -88,9 +88,11 @@ public void beforeDelete(CdsDeleteEventContext context) {
 
 	private void auditCfgChange(final Action action, final ConfigChange cfgChange, EventContext context) {
 		// create new request context and set tenant to null
-		context.getCdsRuntime().requestContext().modifyUser(user -> user.setTenant(null)).run(ctx -> {
-			// send audit log message into provider tenant as user's tenant is null
-			this.auditLog.logConfigChange(action, cfgChange);
+		context.getCdsRuntime().requestContext().modifyUser(user -> user.setTenant(null)).run(requestContext -> {
+			context.getCdsRuntime().changeSetContext().run(changeSetContext -> {
```
mofterdinger, line 92: I don't understand this change. Why is a new ChangeSetContext here required? Should we also adapt the corresponding documentation: https://pages.github.tools.sap/cap/docs/java/auditlog/#subscriber-or-provider-tenant-in-multitenancy-scenario

mta-multi-tenant.yaml
```diff
@@ -47,10 +41,17 @@ modules:
...
+    path: mtx/sidecar
     parameters:
       memory: 256M
       disk-quota: 512M
+    build-parameters:
+      builder: custom
+      build-result: gen
+      commands:
+        - npm run build
+      requires:
```
mofterdinger, line 53: Is this build dependency also required in multi-tenant build? I added it for the single-tenant build, to reuse the installed cds-dk from the previous srv-module build.

mofterdinger: Forget this comment. The sidecar package.json contains just a cds dependency and not cds-dk.

srv/src/main/java/my/bookshop/handlers/AdminServiceAuditHandler.java
beckermarc: This is because I added now tenants to the mock users. The PersistenceService now complains that the tenant changed from "t1" to "null" during the transaction. This is not allowed.
Usually the Outbox detects that in case of MT + tenant "null" the in-memory outbox must be used. However this is not detected here, as the tests run in the single tenant scenario.

This creates a mismatch: PersistenceService behaves as if MT is enabled, while Outbox behaves as MT is disabled. This leads to a conflict, which requires this change to ensure that the log is written in its own transaction. That however goes against the intention of this code. There is no ideal solution for this at the moment.

The best possible change would be to make sure that the tenant is only added to the mock users in case of "local-mtxs". However this is currently also not easily possible, as mock users are maintained in a plain list.

mofterdinger: ok, thanks for clarification.

mofterdinger (approved):

beckermarc:
I wouldn't want to merge this yet. There are still quite a few workarounds/issues in this code that we should solve first:
- @sap/cds-dk installed by maven plugin doesn't suffice for MTXS scenario
- AuditLog workaround necessary, because mock-user configuration can't be split across multiple profiles
- cds deploy triggered during maven build creates empty undefined file in root folder

Let's wait until those issues are resolved from my perspective.

srv/pom.xml
```diff
@@ -173,7 +173,7 @@
 							<goal>npm</goal>
 						</goals>
 						<configuration>
-							<arguments>install</arguments>
+							<arguments>install @sap/cds-dk@${cds.install-cdsdk.version} @sap/cds-mtxs@^1 --no-save</arguments>
```
mofterdinger, line 176: Is this just a workaround or should we improve the goal `install-cdsdk` in a way, that it also installs @sap/cds-mtxs in addition to the @sap/cds-dk? Maybe it can be derived from the project configuration if this dependency needs to be installed or not. Or with an additional switch.

beckermarc: To me all of this should be solved by the @sap/cds-dk tooling itself. For me this is a workaround. Note that the normal @sap/cds-dk would also be fine if installed not from npmjs.org but from artifactory. Because then the node_modules layout is different and the tooling relies on the structure provided when installing from artifactory. We shouldn't built anything of this into the cds-maven-plugin until it is not 100% clear that this is the correct way forward.

mofterdinger: Mhh, really strange that it makes a difference if the cds-dk is installed from npmjs.org or from artifactory. I will never understand how npm works :-(.

PR #195, opened 2023-01-31
Switch to Java 17 as default

vmikhailenko:
Switches sample to Java 17 as default. 

Additional space added to the Java backend to accommodate downloaded JRE. Sidecar is also grew bigger.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,8 +76,7 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.8.1</version>
 				<configuration>
-					<source>${jdk.version}</source>
-					<target>${jdk.version}</target>
+					<release>${jdk.version}</release>
 					<encoding>UTF-8</encoding>
 				</configuration>
 			</plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -15,7 +15,7 @@
 		<revision>1.0.0-SNAPSHOT</revision>
 
 		<!-- DEPENDENCIES VERSION -->
-		<jdk.version>1.8</jdk.version>
+		<jdk.version>17</jdk.version>
 		<cds.services.version>1.31.0</cds.services.version>
 		<spring.boot.version>2.7.8</spring.boot.version>
 		<cloud.sdk.version>4.5.0</cloud.sdk.version>
--- a/values.yaml
+++ b/values.yaml
@@ -32,6 +32,8 @@ srv:
     tag: latest
   env:
     SPRING_PROFILES_ACTIVE: cloud
+    BP_JVM_VERSION: 17
+    BP_JVM_TYPE: JRE
   resources:
     limits:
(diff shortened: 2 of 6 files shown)


pom.xml
```diff
@@ -15,7 +15,7 @@
 		<revision>1.0.0-SNAPSHOT</revision>
 
 		<!-- DEPENDENCIES VERSION -->
-		<jdk.version>1.8</jdk.version>
+		<jdk.version>17</jdk.version>
```
beckermarc, line 18: Please also adapt the Maven Compiler Plugin to use release property instead of source/target property

values.yaml
```diff
@@ -32,6 +32,8 @@ srv:
     tag: latest
   env:
     SPRING_PROFILES_ACTIVE: cloud
+    BP_JVM_VERSION: 17
+    BP_JVM_TYPE: JRE
```
beckermarc, line 36: Is this file relevant during container build time? I would have imagined that the decision which JRE to use is already made when building the containers with `pack`. Does this take these environment variables into account?

vmikhailenko: Hah, you right. It can detect the required Java even without hints. Cool.

beckermarc (approved):

helencorrigan:
Hello @vmikhailenko 
Could you please explain why this upgrade to java 17 was done?

beckermarc:
As of CAP Java `1.31.0`, which was released beginning end of January Java 17 is the recommended Java version. This is documented in our release notes: https://pages.github.tools.sap/cap/docs/releases/jan23#jdk-17-support. To reflect this also in our sample application, we made this change.

helencorrigan:
@beckermarc thank you for this feedback.

helencorrigan:
@beckermarc @vmikhailenko 

Via the community the following issue has been reported : [Error in step 6 of CAP tutorial Reuse a CAP Java Service | SAP Community](https://answers.sap.com/questions/13804742/error-in-step-6-of-cap-tutorial-reuse-a-cap-java-s.html)

Does the SBAS environment need to upgrade?

beckermarc:
> Does the SBAS environment need to upgrade?

Yes I think Java 17 should be available in BAS and easily usable. It might also make sense to make it the default.

helencorrigan:
@beckermarc 
currently we use this SBAS env : https://sap-ux-stage.cry10cf.int.applicationstudio.cloud.sap/index.html
& it has 11, therefore now when we use this repo in SBAS, we get the error reported in the community.

Do you know when the SBAS will be upgraded to 17?

beckermarc:
As said in our internal chat, I'd recommend to get in touch with BAS colleagues. For the time being you can also simply change the `jdk.version` property in `pom.xml` to `11` in your local clone. Then the project will run with Java 11. So far no Java 17 specific features are used.

PR #196, opened 2023-01-31
Kyma Deployment Update

sidakphull:
Switched to Approuter based UI deployment instead of HTML5 Apps based and updated other sections.


.cdsrc.json
```diff
@@ -1,21 +1,32 @@
...
+      },
+      {
+        "for": "hana"
+      }
+    ]
+  },
+  "hana": {
+    "deploy-format": "hdbtable"
+  },
+  "requires": {
```
beckermarc, line 21: `requires` section in `.cdsrc.json` should not be relevant for Java applications. Why is this added?

README.md
```diff
@@ -258,46 +256,6 @@ This project contains a pre-configured configuration file `values.yaml`, you jus
 
 - `<your-container-registry>` - full-qualified hostname of your container registry
 - `domain`- full-qualified domain name used to access applications in your Kyma cluster
-
-#### Use API_BUSSINESS_PARTNER Remote Service (optional)
```
beckermarc, line 262: Why is this removed?

README.md
```diff
@@ -312,47 +270,77 @@ The *Docker Server* is the full qualified hostname of your container registry.
...
-```
+```bash
 pack build $YOUR_CONTAINER_REGISTRY/bookshop-srv \
         --path srv/target/*-exec.jar \
         --buildpack gcr.io/paketo-buildpacks/sap-machine \
         --buildpack gcr.io/paketo-buildpacks/java \
         --builder paketobuildpacks/builder:base \
-        --env SPRING_PROFILES_ACTIVE=cloud
+        --env SPRING_PROFILES_ACTIVE=cloud \
+        --env BP_JVM_VERSION=17
```
beckermarc, line 333: Is this required? Shouldn't this be the default, which is automatically detected by the buildpack from the Maven configuration?

README.md
sidakphull: Support for connectivity requires an update. It is blocked due to [this](https://github.com/SAP/cloud-sdk-backlog/issues/809).

.cdsrc.json
sidakphull: This is required by `cds-dk`. For now, it cannot read the project's features from `application.yaml`.

README.md
sidakphull: The buildpack's defaults changes frequently. It's better if we pass the flag here.

.cdsrc.json
beckermarc: Would it also be possible to pass these information on the command line?
I don't really like to fill the `.cdsrc.json` with information that is just required for CLI calls like `cds add helm`.

README.md
beckermarc: Does this also apply to Java? The issue seems to refer to Cloud SDK Node.js..

sidakphull: I think it does. According to [this](https://github.com/SAP/cloud-sdk-backlog/issues/809#issuecomment-1268050031) comment, some additional information is required to support the `Trusted` mode aka the mode in which Connectivity works in Kyma.

.cdsrc.json
sidakphull: That information isn't just required by `cds add helm`. It is required by the `cds-dk`. You can execute `cds add mta` and see that the configuration added by it is incorrect as well. Yes we can execute `cds add hana` or `cds add xsuaa` to get the correct configuration added for the deployment but these commands corrupt the files in the project. For example:

- `cds add hana` changes the `build` script in `package.json` of `db`.
- `cds add multitenancy` corrupts the `xs-security.json` file.

BTW, now that `streamlined-mtx` branch is merged and `.cdsrc.json`, I see that it is set up for multitenancy only. I think we will need two files here. One for Single Tenant and one for Multi Tenant (similar to mta.yaml files). 

Will it be okay if I create the two files (`.cdsrc-st.json` and `.cdsrc-mt.json`) and place them in a folder `kyma (?)`.  In the instructions, I will then mention that user needs to replace the contents of `.cdsrc.json` in root folder with the contents of the file in `kyma` folder.

beckermarc: > You can execute cds add mta and see that the configuration added by it is incorrect as well.

I guess that is why we added predefined mta.yaml files..

> BTW, now that streamlined-mtx branch is merged and .cdsrc.json, I see that it is set up for multitenancy only

No, at the moment you can still deploy the app as single-tenant, even with the same `.cdsrc.json`. I just verified that again. Why is there a difference required in case of Kyma?

README.md
beckermarc: Okay, I will clarify this also again with the Cloud SDK Java colleagues in our next sync.

.cdsrc.json
sidakphull: When `cds add helm` is executed, `cds-dk` reads the project's configuration and checks which all `features` the project has. After this, it adds the required configuration for each `feature`. 

`cds-dk` uses various sources to check if the project has a specific `feature` or not. One of the source is `.cdsrc.json` file. Now that `.cdsrc.json` file has this specific key `requires.multitenancy` set to `true`, `cds-dk` will think that the project is MultiTenant. So every time `cds add helm` is executed, `cds-dk` will add the multitenant configuration only.

This is why I suggest having two `.cdsrc.json` files.

beckermarc: Hm, I don't really like adding two different `.cdsrc.json` files just for `cds add helm`. Is it possible to specify those `features` on the command line as well, when calling `cds add helm`. This would make us independent of the specific information stored in `.cdsrc.json` or in other places.

sidakphull: not really. The only other option I see is adding pre defined charts for single tenant and multi tenant just like `mta.yaml`.

beckermarc: > The only other option I see is adding pre defined charts for single tenant and multi tenant just like mta.yaml.

That is probably a better option then. However I think this really makes sense with the charts provided by the Unified Runtime, as those are easily "updatable". @davidhunglam made first tests with it and the bookshop in https://github.com/SAP-samples/cloud-cap-samples-java/pull/197. Maybe it makes sense to discuss these topics together in a meeting :)

kyma-st.json
```diff
@@ -0,0 +1,32 @@
...
+        "target": ".",
+        "tasks": [
+            {
+                "for": "java"
+            },
+            {
+                "for": "mtx",
+                "src": ".",
+                "dest": "mtx-sidecar"
+            },
```
beckermarc, line 12:
```suggestion
            {
                "for": "mtx-sidecar"
            },
```

kyma-mt.json
```diff
@@ -0,0 +1,28 @@
+{
+    "requires": {
+        "multitenancy": true,
```
beckermarc, line 3: What about `extensibility` and `toggles`, like in `.cdsrc.json`?

README.md
```diff
@@ -241,57 +241,40 @@ Before you can access the UI using the (tenant-specific) URL to the bookshop(-mt
...
 - Logged into Kyma Runtime (with `kubectl` CLI), Cloud Foundry space (with `cf` CLI) and Container Registry (with `docker login`)
-- `@sap/cds-dk` >= 6.0.1
+- `@sap/cds-dk` >= 6.6.0
 
 ### Add Deployment Files
 
-CAP tooling provides your a Helm chart for deployment to Kyma.
+CAP tooling provides you a Helm chart for deployment to Kyma.
+
+For single tenant deployment, replace the contents of _`cdsrc.json`_ with _`kyma-st.json`_.
```
beckermarc, line 250:
```suggestion
For single tenant deployment, replace the contents of _`.cdsrc.json`_ with _`kyma-st.json`_.
```

README.md
```diff
@@ -241,57 +241,40 @@ Before you can access the UI using the (tenant-specific) URL to the bookshop(-mt
...
+For multi tenant deployment, replace the contents of _`cdsrc.json`_ with _`kyma-mt.json`_.
```
beckermarc, line 251:
```suggestion
For multi tenant deployment, replace the contents of _`.cdsrc.json`_ with _`kyma-mt.json`_.
```

README.md
```diff
@@ -300,6 +283,8 @@ You can try the `API_BUSINESS_PARTNER` service with a real S/4HANA system with t
     cds add connectivity
     ```
 
+5. `cds add helm` will not add configuration required to create a Connectivity Service Instance. This Service Instance should be created by the Kyma Cluster Administrator. For more information regarding configuration of Connectivity Instance, please check the [documentation](https://cap.cloud.sap/docs/guides/deployment/deploy-to-kyma#connectivity-service).
```
beckermarc, line 286: Should this rather be part of step 4. A connectivity service instance seems to be only required for on-premise

README.md
```diff
@@ -314,78 +299,196 @@ bash ./scripts/create-container-registry-secret.sh
...
-### Build
+```
+bash ./scripts/create-sm-secret.sh bookshop-sm
+```
+
+It will create a HDI container `bookshop-db` (single tenant) or a Service Manager `bookshop-sm` (multi tenant) instance on your currently targeted Cloud Foundry space and creates a secret `bookshop-db` (single tenant) or `bookshop-sm` (multi tenant) with the credentials in your current Kubernetes namespace.
+
+Make the following changes to your _`chart/values.yaml`_.
+
+##### Single Tenant
```
beckermarc, line 322: What about combining both multi tenant and single tenant sections into one to only have 2 headings instead of 4?

README.md
```diff
@@ -314,78 +299,196 @@ bash ./scripts/create-container-registry-secret.sh
...
+```bash
+helm install bookshop ./chart --set-file xsuaa.jsonParameters=xs-security.json
+```
 
 Before you can access the UI you should make sure to [Setup Authorizations in SAP Business Technology Platform](#setup-authorizations-in-sap-business-technology-platform).
 
-1. Create a Launchpad Service subscription in the BTP Cockpit
-2. Go to **HTML5 Applications**
-3. Start any of the HTML5 applications
+Click on the approuter url logged by the `helm upgrade` to access the UI.
```
beckermarc, line 483: `helm upgrade` or `helm install`?

README.md
sidakphull: done

README.md
sidakphull: yeah. I added it as a Note to step 4.

README.md
```diff
@@ -314,78 +299,198 @@ bash ./scripts/create-container-registry-secret.sh
...
+```bash
+pack build $YOUR_CONTAINER_REGISTRY/bookshop-approuter \
+     --path app \
+     --buildpack gcr.io/paketo-buildpacks/nodejs \
+     --builder paketobuildpacks/builder:base \
+     --env BP_NODE_RUN_SCRIPTS=""
 ```
 
-**Build HTML5 application deployer image:**
+**Build data base deployer image (single tenant only):**
```
beckermarc, line 400:
```suggestion
**Build database deployer image (single tenant only):**
```

beckermarc (approved):
I made a few more adaptions. I thought it was nicer when we had the instructions what to change in .cdsrc.json directly in README and avoid introducing two new files that need to be maintained as well, when we make changes outside of the requires section of .cdsrc.json.

Hope the changes are fine with you. From my perspective this can be merged now :)

sidakphull:
The changes look good. Thanks @beckermarc!

PR #206, opened 2023-02-15
Fix `cds add helm` commands

swaldmann:
The `cds add helm:<feature>` commands don't exist any more. They are now merged with the standard `cds add <feature>` commands.

@BraunMatthias @mofterdinger @beckermarc


sidakphull:
Fixed in #196 along with a few other things.

BraunMatthias (approved):

PR #223, opened 2023-05-02
Use CDS Services 1.34.0

beckermarc:


beckermarc:
Yes, the local build works for me. I guess this is an inconsistency in your setup, as the tests also run in the GitHub pipeline.

mofterdinger (approved):

PR #228, opened 2023-05-12
Use npm workspaces to reuse @sap/cds-mtxs from MTX Sidecar

beckermarc:
npm workspaces can help to share @sap/cds-mtxs from MTX Sidecar with the version used by @sap/cds-dk and cds build.
Main downside at the moment: To retain fixed versions when using @sap/cds-dk this requires a package-lock.json in root.
In addition the package-lock.json in mtx/sidecar is required for cds build, as only this file is copied over to gen.

### DIFF
--- a/.cdsrc.json
+++ b/.cdsrc.json
@@ -5,20 +5,10 @@
         "toggles": true
     },
     "build": {
-        "target": ".",
-        "tasks": [
-            {
-                "for": "java"
-            },
-            {
-                "for": "mtx-sidecar"
-            },
-            {
-                "for": "hana"
-            }
-        ]
+        "target": "."
     },
-    "hana": {
-        "deploy-format": "hdbtable"
-    }
+    "profiles": [
+        "with-mtx-sidecar",
+        "java"
+    ]
 }
--- a/.gitignore
+++ b/.gitignore
@@ -20,7 +20,6 @@ target/
 
 node/
 node_modules/
-package-lock.json
 
 .mta/
 *.mtar
--- a/integration-tests/pom.xml
+++ b/integration-tests/pom.xml
@@ -75,6 +75,20 @@
 				<artifactId>cds-maven-plugin</artifactId>
 				<version>${cds.services.version}</version>
 				<executions>
+					<!-- Following steps start and stop MTX sidecar for tests executed by failsafe plugin -->
+					<execution>
+						<id>install-sidecar</id>
+						<goals>
+							<goal>npm</goal>
+						</goals>
+						<phase>pre-integration-test</phase>
+						<configuration>
+							<arguments>install</arguments>
+							<skip>${skipTests}</skip>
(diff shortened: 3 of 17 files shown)


beckermarc:
Follow up of https://github.com/SAP-samples/cloud-cap-samples-java/pull/224
FYI @LotharBender @swaldmann @mofterdinger

LotharBender:
I've just noticed that the `cds-maven-plugin` does not check whether the installed @sap/cds-dk version matches the configured version - the configured version is not installed.
In the node_modules folder @sap/cds-dk@6.7.2 is installed. After updating `<cds.install-cdsdk.version>6.8.0</cds.install-cdsdk.version>` the command `mvn clean install` logs:
> [INFO] InstallCdsdkMojo: Found locally installed @sap/cds-dk, skipping execution.

After deleting the `node_modules` folder and executing `mvn clean install` again, the @sap/cds-dk version is installed correctly:
> [INFO] InstallCdsdkMojo: Executing [/Users/d022960/.m2/repository/com/sap/cds/cds-maven-plugin/cache/unpacked/16.19.0/unpacked-16.19.0-darwin-x64.tar.gz/node-v16.19.0-darwin-x64/bin/npm, install, @sap/cds-dk@6.8.0, --no-save] in working directory /Users/d022960/git/work/java-mtx2

mofterdinger:
> I've just noticed that the cds-maven-plugin does not check whether the installed @sap/cds-dk version matches the configured version - the configured version is not installed.

@LotharBender Hi Lothar, yes this behavior is described here: https://cap.cloud.sap/docs/java/assets/cds-maven-plugin-site/install-cdsdk-mojo.html

By default, this goal looks for an already installed @sap/cds-dk and skips installation if it was found. It doesn't validate the found version against the requested version and the existing @sap/cds-dk could be outdated. Add property -Dcds.install-cdsdk.force=true to the Maven command line to force the installation of the @sap/cds-dk in the configured version.

LotharBender:
> By default, this goal looks for an already installed @sap/cds-dk and skips installation if it was found. It doesn't validate the found version against the requested version and the existing @sap/cds-dk could be outdated. 

But doesn't this again prove that Java projects should better stick to the npm approach of maintaining and updating npm dependencies instead of using a maven based mechanism?
- outdated local cds-dk versions have caused a lot of problems in the past
- npm would correctly update the cds-dk version by default. A package-lock can be used to install fixed versions for @sap/mtxs and @sap/cds-dk.
- a new version would only be downloaded if required. I assume that `cds.install-force=true` will always download and install a new version, regardless whether the installed version already matches or not. Because of the download size of the cds-dk this would be very expensive.

On the other hand this means that the default local development behaviour based on the current cds-dk installation approach is no way better than using `npm workspaces` without a package-lock.json in project root - I would say it's worse ;-)

beckermarc:
Newer proposal: https://github.com/SAP-samples/cloud-cap-samples-java/pull/346

PR #239, opened 2023-07-13
Remove `pm` dev dependency from sidecar `package.json`

swaldmann:
Hi @beckermarc, it looks like the `pm` dependency is not used, so it shouldn't be required here.

### DIFF
--- a/mtx/sidecar/package.json
+++ b/mtx/sidecar/package.json
@@ -8,8 +8,7 @@
     "passport": "^0"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


beckermarc:
It is used here: https://github.com/SAP-samples/cloud-cap-samples-java/blob/main/integration-tests/pom.xml#L98.
However npx I guess installs it also on the fly over there. Not sure, if it would be more explicit though if it is explicitly declared in the package.json.

swaldmann:
I see, I think we can leave it in then.

PR #244, opened 2023-08-01
PostgreSQL-based bookshop

vmikhailenko:
Runnable (`mvn spring-boot:test-run` and from the IDE) and deployable (BTP-only) bookshop with PostgreSQL as the database. I adjusted the setup, removed everything that mocked S/4 HANA and rest of the stuff.


beckermarc:
Should we minimize the sample even further?
- Minimize db CDS artifacts
- Minimize srv CDS artifacts + custom code
- Remove or minimize UI?

vmikhailenko:
> Should we minimize the sample even further?
> 
> * Minimize db CDS artifacts
> * Minimize srv CDS artifacts + custom code
> * Remove or minimize UI?

I would vote to remove Vue and Swagger. If the goal is to make the sample as slim as possible, we can leave only Books/Authors and administration of books to have read and write scenario.

mofterdinger:
> I would vote to remove Vue and Swagger.

+1

pom.xml
```diff
@@ -18,15 +18,14 @@
 		<jdk.version>17</jdk.version>
 		<cds.services.version>2.1.0</cds.services.version>
 		<spring.boot.version>3.1.2</spring.boot.version>
-		<cloud.sdk.version>4.18.0</cloud.sdk.version>
-		<xsuaa.version>3.1.0</xsuaa.version>
+		<xsuaa.version>3.0.1</xsuaa.version>
```
beckermarc, line 21: Let's remove XSUAA dependency management. It was mainly introduced to ensure consistency with Cloud SDK, but this was also removed from this project.

pom.xml
```diff
@@ -18,15 +18,14 @@
...
 		<cf-java-logging-support.version>3.7.0</cf-java-logging-support.version>
-		<cds.install-cdsdk.version>7.0.3</cds.install-cdsdk.version>
+		<liquibase.version>4.20.0</liquibase.version>
+		<testcontainers.version>1.18.3</testcontainers.version>
```
beckermarc, line 24: Liquibase and Testcontainers don't need to be managed, as they come with the Spring Boot dependencies BOM. We should use versions from there.

srv/admin-service.cds
```diff
@@ -1,38 +1,18 @@
...
-@path : 'admin'
-service AdminService @(requires : 'admin') {
-  entity Books   as projection on my.Books actions {
-    action addToOrder(order_ID : UUID, quantity : Integer) returns Orders;
+@path: 'admin'
+service AdminService @(requires: 'admin') {
+  @cds.search: {
+    descr,
+    title
   }
```
beckermarc, line 9: Why was this annotation added?

vmikhailenko: From here https://github.com/SAP-samples/cloud-cap-samples-java/blob/5371e0532fda61983a73267eea08bfbeb2da2627/srv/admin-service.cds#L27

srv/src/test/java/my/bookshop/AdminServiceTest.java
```diff
@@ -1,100 +1,57 @@
 package my.bookshop;
```
beckermarc, line 1: Let's delete this test and CatalogServiceTest and just keep SimpleIntegrationTest.

srv/src/main/resources/application.yaml
```diff
@@ -1,27 +1,26 @@
 ---
-logging:
-  level:
-    '[com.sap.cds.auditlog]': DEBUG
+logging.level:
+  postgres: INFO
```
beckermarc, line 3: Isn't INFO the default? Remove it?

mta.yaml
```diff
@@ -62,32 +48,36 @@ modules:
...
     requires:
       - name: app-api
-  - name: bookshop-hdi-container
+
+  - name: bookshop-pg-db
     type: org.cloudfoundry.managed-service
     parameters:
-      service: hana
-      service-plan: hdi-shared
+      path: ./postgresql-db-config.json
```
beckermarc, line 77: Let's put the configuration directly in here under `config`, instead of separating this single line into a dedicated file.

pom.xml
```diff
@@ -17,16 +17,13 @@
 		<!-- DEPENDENCIES VERSION -->
 		<jdk.version>17</jdk.version>
 		<cds.services.version>2.1.0</cds.services.version>
+		<xsuaa.version>3.0.1</xsuaa.version>
```
beckermarc, line 20: Maybe we should rather use the `cds-starter-cloudfoundry` in srv/pom.xml which also includes all required XSUAA dependencies with an exclusion for the cds-feature-hana and cds-feature-mt? What do you think? This would allow us to use the XSUAA version coming from CAP.

vmikhailenko: Ah, dumb me. Thanks!

PR #252, opened 2023-10-06
Switch from sap_java_buildpack to java_buildpack

vmikhailenko:
We need to switch the Postgres-based sample from the SAP Java Buildpack to the CF one. SAP buildpack does not work with SSL support we delivered in the recent release.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
 
 		<!-- DEPENDENCIES VERSION -->
 		<jdk.version>17</jdk.version>
-		<cds.services.version>2.2.0</cds.services.version>
+		<cds.services.version>2.3.0</cds.services.version>
 		<spring.boot.version>3.1.3</spring.boot.version>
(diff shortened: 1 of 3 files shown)


mofterdinger:
@vmikhailenko Hi, what's missing here?

vmikhailenko:
Nothing. outbound-tests run as far as I can see (there is an PR there). If no one has objections, we can merge this.

mofterdinger (approved):

pom.xml
```diff
@@ -16,7 +16,7 @@
 
 		<!-- DEPENDENCIES VERSION -->
 		<jdk.version>17</jdk.version>
-		<cds.services.version>2.2.0</cds.services.version>
+		<cds.services.version>2.3.0</cds.services.version>
 		<spring.boot.version>3.1.3</spring.boot.version>
```
beckermarc, line 20:
```suggestion
		<spring.boot.version>3.1.4</spring.boot.version>
```

beckermarc (approved):

PR #257, opened 2023-10-23
Use bundled SAPMachine version from sap_java_buildpack

vmikhailenko:
It seems that we can use the SAPMachine 17 bundled into `sap_java_buildpack`.


mofterdinger:
@vmikhailenko Hi Vladimir, did you verify this change with a deployment to BTP ?

vmikhailenko:
> @vmikhailenko Hi Vladimir, did you verify this change with a deployment to BTP ?

Yes. There is a PR in outbound tests to verify this.

beckermarc:
Is this also available already on Live Landscapes, e.g. EU10?

vmikhailenko:
> Is this also available already on Live Landscapes, e.g. EU10?

Probably. The initial support was released in August. The ` cf buildpacks | grep java` shows me 1.78.0 there.

PR #258, opened 2023-10-24
Postgres version upgrades

rjayasinghe:


pom.xml
```diff
@@ -19,6 +19,7 @@
 		<cds.services.version>2.3.0</cds.services.version>
 		<spring.boot.version>3.1.4</spring.boot.version>
 		<cf-java-logging-support.version>3.7.0</cf-java-logging-support.version>
+		<cds.install-node.nodeVersion>v18.18.2</cds.install-node.nodeVersion>
```
beckermarc, line 22: Node.js 18 will be the default with the upcoming 2.4.0 release in our CDS Maven Plugin. If we add this now, we should not forget to remove it again later, as we would want to rely on the defaults..

mofterdinger: yes, at the moment the build fails and we need to configure the Node.js version.

PR #262, opened 2023-10-27
Simplify `.cdsrc.json`

swaldmann:
With CDS 7…

- The `target: "."` is automatically coming with profile `java`
- The `java` profile is auto-assigned if a `pom.xml` is found

→ We can simplify the config here even further

@beckermarc


beckermarc (approved):
Thanks :)

PR #265, opened 2023-11-08
configure cds-dk version in pom.xml

mofterdinger:


pom.xml
```diff
@@ -16,10 +16,12 @@
...
 		<!-- DEPENDENCIES VERSION -->
 		<jdk.version>17</jdk.version>
-		<cds.services.version>2.3.0</cds.services.version>
-		<spring.boot.version>3.1.4</spring.boot.version>
+		<cds.services.version>2.3.1</cds.services.version>
+		<spring.boot.version>3.1.5</spring.boot.version>
 		<cf-java-logging-support.version>3.7.0</cf-java-logging-support.version>
 		<cds.install-node.nodeVersion>v18.18.2</cds.install-node.nodeVersion>
+
+		<cds.cdsdk-version>7.3.1</cds.cdsdk-version>
```
beckermarc, line 24: I think this feature will only be released with `2.4.0`

mofterdinger:
I wonder why the build works with cds-maven-plugin 2.3.1, because support for the new property `<cds.cdsdk-version>` comes with 2.4.0.

pom.xml
mofterdinger: yes, I know, but the maven build works, really strange. We update this PR to use 2.4.0 once it's released.

mofterdinger: Yes, I found it also in the package.json of @cap-js/postgres. I wonder if we could use this peer dependency to install the cds-dk, because it ensures that the @cap-js/postgres has the correct cds-dk installed. If we configure our own version of the cds-dk for the postgres setup, it could be incompatible.

beckermarc: Not sure, if this is how it is intended to be used. I guess you want to have the flexibility to update cds-dk independently in case of bug fixes and so on.

mofterdinger: With version 2.4.0 of cds-services it's possible to use the new approach to install cds-dk with npx option --package.

mofterdinger:
Hi, this PR is now ready for review. Thanks, Markus

beckermarc (approved):

PR #266, opened 2023-11-14
Update to cds-services 2.4.0

StefanHenke:


pom.xml
```diff
@@ -23,7 +23,7 @@
 
 		<!-- DEPENDENCIES VERSION -->
 		<jdk.version>17</jdk.version>
-		<cds.services.version>2.3.1</cds.services.version>
+		<cds.services.version>2.4.0</cds.services.version>
 		<cloud.sdk.version>4.26.0</cloud.sdk.version>
 		<xsuaa.version>3.1.3</xsuaa.version>
```
beckermarc, line 28: Please also bump other dependencies maintained here to the recommended version, we use with 2.4.0 -> e.g. XSUAA version was updated to `3.2.1` with `2.4.0`

StefanHenke: There is a new version `3.8.0` of the cf-java-logging-support library since yesterday. I kept it with `3.7.1` as we haven´t done any testing yet with it.

beckermarc:
Should we also adopt the new `systemUserProvider()` method with this change here?
https://github.com/SAP-samples/cloud-cap-samples-java/blob/4635ca8f1b9474d1891cc82857d9d974e4172542/srv/src/main/java/my/bookshop/handlers/AdminServiceAuditHandler.java#L91

beckermarc (approved):

StefanHenke:
> Should we also adopt the new `systemUserProvider()` method with this change here?
> 
> https://github.com/SAP-samples/cloud-cap-samples-java/blob/4635ca8f1b9474d1891cc82857d9d974e4172542/srv/src/main/java/my/bookshop/handlers/AdminServiceAuditHandler.java#L91

Makes perfectly sense

PR #271, opened 2023-11-22
Update @sap/cds-dk to 7.4.0

beckermarc:
Disable generation of localized EDMX.


mofterdinger (approved):
Should we add this new option also to the cds-services-archetype ?

PR #274, opened 2023-11-29
Use Node.js 18, 16 no longer available at BTP

mofterdinger:


StefanHenke:
There is at least one more occurrence of node 16 in https://github.com/SAP-samples/cloud-cap-samples-java/blob/main/db/package.json. I´m not sure if it would block deployment if we stay on 16, but I think it is more consistent to go to 18 as well. Node 16 in general seems still available, but reached its EOL already.

StefanHenke (approved):

PR #279, opened 2023-12-18
Update CDS Services to 2.5.0

beckermarc:


StefanHenke (approved):
we should not forget the other branches. PR #280 addresses at least postgres

PR #294, opened 2024-02-27
Added postgres branch to build trigger

mofterdinger:
- added postgres branch to build trigger
- updated actions to latest versions

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -24,5 +24,6 @@ jobs:
       uses: actions/setup-java@v4
       with:
         java-version: 17
+        distribution: 'temurin'
     - name: Build with Maven
       run: mvn -B clean install
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -5,9 +5,13 @@ name: Java CI with Maven
 
 on:
   push:
-    branches: [ main ]
+    branches:
+      - main
+      - postgres
   pull_request:
-    branches: [ main ]
+    branches: 
+      - main
+      - postgres
 
 jobs:
   build:
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -19,11 +19,16 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/maven.yml
```diff
@@ -5,20 +5,30 @@ name: Java CI with Maven
...
     steps:
-    - uses: actions/checkout@v2
+    - name: Checkout
+      uses: actions/checkout@v4
     - name: Set up JDK
-      uses: actions/setup-java@v1
+      uses: actions/setup-java@v4
       with:
         java-version: 17
+        distribution: 'temurin'
```
beckermarc, line 28: Could we specify SAP Machine here?

mofterdinger: no, not available: https://github.com/actions/setup-java?tab=readme-ov-file#supported-distributions
I checked it already, I would also prefer to use SAPMachine

.github/workflows/maven.yml
```diff
@@ -5,20 +5,30 @@ name: Java CI with Maven
...
+    - name: Setup Maven
+      uses: stCarolas/setup-maven@v4.5
+      with:
+        maven-version: 3.9.6
```
beckermarc, line 32: With which version did we run earlier? `stCarolas` sounds a little inofficial. Hopefully this is trustworthy ;)

mofterdinger: > With which version did we run earlier?

To be honest I don't know, I guess it uses the maven version that is pre-installed in the container of the github runner.

> stCarolas sounds a little inofficial. Hopefully this is trustworthy ;)

it's listed here: https://github.com/marketplace/actions/setup-maven-action
I'm using it already for a couple of month in a private project without any problmes.

beckermarc: Okay, I am wondering if the additional Maven setup build times are worth it. I guess this involves some download and so on. If the version we are using at the moment also works fine :)

mofterdinger: ok, will remove it again. It was just some kind of preparation for the next major release when we bump the min. maven version. Then we can add it if required.

beckermarc (approved):

PR #296, opened 2024-03-06
Use cds-dk 7.7.1

mofterdinger:
PostgreSQL 14 is available in BTP: https://help.sap.com/docs/postgresql-hyperscaler-option/postgresql-on-sap-btp-hyperscaler-option-c92112ee69784c3383a0fb8361156a6f/2023-what-s-new-for-postgresql-hyperscaler-option-archive?locale=en-US

Parameters overview: https://help.sap.com/docs/postgresql-hyperscaler-option/postgresql-on-sap-btp-hyperscaler-option-c92112ee69784c3383a0fb8361156a6f/parameters?locale=en-US

### DIFF
--- a/mta.yaml
+++ b/mta.yaml
@@ -77,7 +77,7 @@ resources:
       service: postgresql-db
       service-plan: standard
       config:
-        engine_version: "13"
+        engine_version: "14"
         multi_az: false
 
   - name: cf-logging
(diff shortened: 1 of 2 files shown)


vmikhailenko (approved):

vmikhailenko:
I will trigger the outbound test once. We may have to re-create Postgres instance that we use there.

mofterdinger:
@vmikhailenko Yes, the update is already done. I wonder if we could also use PostgreSQL 15, at least in the engine versions value list it's available. Default is 14, so it's safer to use this version I think

PR #297, opened 2024-03-06
Delete githbu workflow from postgres branch

mofterdinger:
Remove github workflow from postgres branch, only considered by main branch.


mofterdinger:
Mhh, looks like my assumption is wrong and it's also required in the postgres branch, no build is running for this PR.
Last week I added the postgres branch to the workflow in main branch to get a build triggered.

PR #305, opened 2024-04-04
Technical Provider User should run with all feature toggles

beckermarc:


srv/src/main/java/my/bookshop/config/CustomFeatureToggleProvider.java
```diff
@@ -17,6 +17,11 @@ public class CustomFeatureToggleProvider implements FeatureTogglesInfoProvider {
 
 	@Override
 	public FeatureTogglesInfo get(UserInfo userInfo, ParameterInfo parameterInfo) {
+		if (userInfo.getTenant() == null && userInfo.isSystemUser()) {
+			// technical provier user runs with all feature toggles
```
StefanHenke, line 21:
```suggestion
			// technical provider user runs with all feature toggles
```

StefanHenke (approved):

PR #314, opened 2024-04-25
pm2-less integration tests

vmikhailenko:
`pm2` has very confusing license. Let's try to get rid of it.

### DIFF
--- a/integration-tests/pom.xml
+++ b/integration-tests/pom.xml
@@ -95,34 +95,32 @@
 							<workingDirectory>${sidecar.dir}</workingDirectory>
 						</configuration>
 					</execution>
-
-					<execution>
-						<id>start-sidecar</id>
-						<goals>
-							<goal>npx</goal>
-						</goals>
-						<phase>pre-integration-test</phase>
-						<configuration>
-							<arguments>pm2 -l sidecar.log --time start --name=sidecar npm -- run start</arguments>
-							<skip>${skipTests}</skip>
-							<workingDirectory>${sidecar.dir}</workingDirectory>
-						</configuration>
-					</execution>
-
-					<execution>
-						<id>stop-sidecar</id>
-						<goals>
-							<goal>npx</goal>
-						</goals>
-						<phase>post-integration-test</phase>
-						<configuration>
-							<arguments>pm2 delete sidecar</arguments>
-							<skip>${skipTests}</skip>
-							<workingDirectory>${sidecar.dir}</workingDirectory>
-						</configuration>
-					</execution>
 				</executions>
 			</plugin>
+
+			<plugin>
+				<groupId>org.codehaus.mojo</groupId>
+				<artifactId>exec-maven-plugin</artifactId>
+				<version>3.2.0</version>
+				<executions>
+					<execution>
+						<phase>pre-integration-test</phase>
+						<id>start-sidecar</id>
+						<goals>
+							<goal>exec</goal>
+						</goals>
+					</execution>
+				</executions>
+				<configuration>
+					<executable>node</executable>
+					<!-- optional -->
+					<workingDirectory>${sidecar.dir}</workingDirectory>
+					<async>true</async>
+					<useMavenLogger>true</useMavenLogger>
+					<commandlineArgs>node_modules/@sap/cds/bin/cds-serve.js</commandlineArgs>
+				</configuration>
+			</plugin>
+
 		</plugins>
 	</build>
 </project>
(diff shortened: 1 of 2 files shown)


mofterdinger:
> pm2 has very confusing license

What license is it using? Has anyone from legal contacted you about a licensing issue or how did it come about?

integration-tests/pom.xml
```diff
@@ -95,33 +95,29 @@
...
-						<phase>post-integration-test</phase>
-						<configuration>
-							<arguments>pm2 delete sidecar</arguments>
-							<skip>${skipTests}</skip>
-							<workingDirectory>${sidecar.dir}</workingDirectory>
-						</configuration>
 					</execution>
 				</executions>
+				<configuration>
+					<executable>node</executable>
```
beckermarc, line 115: Does this work if the node executable is not available in the path? Would it find the node.js installation from the CDS Maven Plugin?

integration-tests/pom.xml
```diff
@@ -95,33 +95,29 @@
...
+				<executions>
 					<execution>
-						<id>start-sidecar</id>
-						<goals>
-							<goal>npx</goal>
-						</goals>
 						<phase>pre-integration-test</phase>
-						<configuration>
-							<arguments>pm2 -l sidecar.log --time start --name=sidecar npm -- run start</arguments>
-							<skip>${skipTests}</skip>
```
beckermarc, line 107: Could we retain the skip configuration, so that MTX Sidecar is not started, in case tests are skipped? Or would this anyway be the case, because phase `pre-integration-test` is never executed?

integration-tests/pom.xml
mofterdinger: You can inject the installation path of the node binary, see here: https://cap.cloud.sap/docs/java/assets/cds-maven-plugin-site/install-node-mojo.html
${cds.node.executable}: Location of the node executable

integration-tests/pom.xml
```diff
@@ -82,46 +82,53 @@
...
-						<phase>pre-integration-test</phase>
 						<configuration>
-							<arguments>install --no-save</arguments>
-							<skip>${skipTests}</skip>
-							<workingDirectory>${sidecar.dir}</workingDirectory>
+							<skip>${cdsdk-global}</skip>
 						</configuration>
 					</execution>
 
+					<!-- Following steps start and stop MTX sidecar for tests executed by failsafe plugin -->
```
beckermarc, line 95: Comment is a little outdated here.. maybe just remove completely..

integration-tests/pom.xml
```diff
@@ -82,46 +82,53 @@
...
 						<goals>
-							<goal>npm</goal>
+							<goal>install-node</goal>
 						</goals>
-						<phase>pre-integration-test</phase>
 						<configuration>
-							<arguments>install --no-save</arguments>
-							<skip>${skipTests}</skip>
-							<workingDirectory>${sidecar.dir}</workingDirectory>
+							<skip>${cdsdk-global}</skip>
```
beckermarc, line 91: I think `${cds.node.executable}` is not set, if the `install-node` goal is skipped. Would be great if this property would default to the executable name `node`, assuming that if `install-node` is not executed it is available in the path.
@mofterdinger Am I correct? What do you think?

integration-tests/pom.xml
```diff
@@ -95,33 +95,29 @@
...
-							<workingDirectory>${sidecar.dir}</workingDirectory>
-						</configuration>
 					</execution>
 				</executions>
+				<configuration>
+					<executable>node</executable>
+					<workingDirectory>${sidecar.dir}</workingDirectory>
+					<async>true</async>
+					<useMavenLogger>true</useMavenLogger>
+					<commandlineArgs>node_modules/@sap/cds/bin/cds-serve.js</commandlineArgs>
```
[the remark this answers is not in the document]
beckermarc: Would this work?

integration-tests/pom.xml
mofterdinger: Yes, you are right, if the goal `install-node` is not executed, the property is not set.
We could try to set the value of the property in the pom.xml first: <cds.node.executable>node</cds.node.executable>.
If the install-node is execute, it will overwrite this value.

mofterdinger: I'm working on a PR to set just the executable names node/npm/npx in case the goal is skipped.

beckermarc (approved):

PR #315, opened 2024-04-29
Updated to cds-services 2.9.0

BraunMatthias:
Removed the custom index page UI links, as the UI is now served and added to the index page automatically.


beckermarc (changes requested):
We must also remove the old index page UI links. I will add this to this change.

PR #323, opened 2024-05-24
Move `cfenv` switch for PostgreSQL  to `application yaml`

vmikhailenko:
Perhaps, `application.yaml` is a better place for the switch in question.


StefanHenke (approved):

vmikhailenko:
Reason: the service instance name is rather property of the deployment configuration and not the application properties.

PR #325, opened 2024-05-28
Showcase annotation expression

eugene-andreev:
Demonstrate the usage of [Expressions as Annotation Values](https://cap.cloud.sap/docs/cds/cdl#expressions-as-annotation-values) to calculate the discounted books' title.

### DIFF
--- a/srv/cat-service.cds
+++ b/srv/cat-service.cds
@@ -31,6 +31,10 @@ service CatalogService @(requires: 'any') {
             to : 'authenticated-user'
         }
     ];
+    
+    extend projection Books with {
+        virtual null as percentDiscount : Integer
+    }
 
-    annotate Books : title with @discountTitlePattern: (title + '-- %d%% discount');
+    annotate Books : title with @discountTitlePattern: (title || ' -- ' || percentDiscount || '% discount');
 }
--- a/srv/cat-service.cds
+++ b/srv/cat-service.cds
@@ -32,5 +32,5 @@ service CatalogService @(requires: 'any') {
         }
     ];
 
-    annotate Books : title with @discountTitlePattern: (title + ' -- %d%% discount');
+    annotate Books : title with @discountTitlePattern: (title + '-- %d%% discount');
 }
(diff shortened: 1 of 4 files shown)


agoerler (commented):
I think this example raises many questions related to the expression `(title + '-- %d%% discount')`:

* why are you using the operator `+` for concatenation and not `||`?
* why is the concat operator `||` not standardized in CAP Java or CAP?
* why is `title` a ref but `discount` isn't?
* `title` is an element of `Books` but `discount` isn't. It's rather a parameter. Can we use a parameter in an annotation expression?
* Should the expression be something like `(title || ' -- %d%% ' || :discount')` ?
* Do we represent the expression as expected in the token tree?

I think the representation should be something like 

plain:||
  |     - param:discount
plain:||
  |     - ' -- %d%% '
ref:title


I think there is too much unclear to showcase this as of now

eugene-andreev:
> I think this example raises many questions related to the expression `(title + '-- %d%% discount')`:

> * `title` is an element of `Books` but `discount` isn't. It's rather a parameter. Can we use a parameter in an annotation expression?

Not sure how to use parameters in CDS, although there is a description for CSN https://cap.cloud.sap/docs/cds/cxn#binding-parameters

Using virtual field instead.

beckermarc:
Hm my first impression is that this is a slightly weird/artificial use case to show this. Would you actually do it like that? What are the benefits? Seems like the code is much more complicated without gaining much. Maybe we should find a better example or am I missing something?

eugene-andreev:
> Hm my first impression is that this is a slightly weird/artificial use case to show this. Would you actually do it like that? What are the benefits? Seems like the code is much more complicated without gaining much. Maybe we should find a better example or am I missing something?

This is an "advanced" way to implement/evaluate ANY free-flow expression.
CAP Java doesn't provide a way to evaluate expressions (be it arithmetic or string xpr) out-of-the-box. I've also mentioned this missing feature in the BLI.
If we manage to introduce expression calculator/evaluator/interpreter as a separate component, this may simplify the whole logic and get rid of custom implementation of `CqnVisitor`. However, so far, this is not the case.

agoerler:
> to concat strings: `title || discount` one would need to implement own logic.

We could introduce a concat operator with corresponding CQN node. Only then we can process concat reasonably with a vistor.

PR #336, opened 2024-06-26
doc(README): Make shell snippet copy&paste safe

bugwelle:
Many bash/shell snippets in the README use an environment variable `YOUR_CONTAINER_REGISTRY`, which is used as `$YOUR_CONTAINER_REGISTRY`.

However, that has the downside that if users copy & paste the snippet and accidentally press enter or use a shell that doesn't allow multi- line pasting (and executes it immediately), then the variable will be replaced by an empty string, because it wasn't set before.

It's good practice to make those snippets copy&paste-safe. We can achieve that by adding `:?` to the variable, which emits an error if the variable is unset or null.

See https://devhints.io/bash for a bash cheat sheet.

Example:

```sh
docker push ${YOUR_CONTAINER_REGISTRY:?}/bookshop-srv
```

On ZSH:

```
zsh: YOUR_CONTAINER_REGISTRY: parameter not set
```

On Bash:

```
bash: YOUR_CONTAINER_REGISTRY: parameter null or not set
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ Use the following steps to import the project to Eclipse:
 
 	```${workspace_loc:bookshop-parent}```
 
-	Afterwards, click **Run**. This step starts the applications `main` method located in `src/main/java/my/bookshop/Application.java`.
+	Afterward, click **Run**. This step starts the applications `main` method located in `src/main/java/my/bookshop/Application.java`.
 
 ## Using IntelliJ Idea (Community and Ultimate)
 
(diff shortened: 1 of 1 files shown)


davidhunglam (approved):

davidhunglam:
Actually most people replace the whole string in the copied command. But in cases where an actual env variable is used, this is quite nice.

bugwelle:
@davidhunglam The issue is that some shells auto-executed a pasted snippet if it ends with a trailing newline. That has happened a few times in the past for me, when I didn't setup my shell properly. 😄 

This is just a safe-guard. I'd also expect users to replace the variable altogether and not set it in their shell.

bugwelle:
Thanks for reviewing my change, David! 😄 
I don't have merge-permissions. Unless anybody else needs to review this, it would be great if you could merge this PR. Thanks!

PR #345, opened 2024-07-19
Upgrade all components to cds8

StefanHenke:
This PR fixes test failures of the Bookshop ITests using CAP 3 and cds8.

* `authentication.normalize-provider-tenant: true` needs to be set in the `application.yaml` because current itest logic requires the provider tenant to be subscribed. The deployment shell script is setting this value to `false` using a `sed` command which requires this property to be available in the file
* `hdi-deploy` in version `5` does not bundle any hana db client, but comes with a post install script that checks that one client has been installed by an external party. As we don´t need db accesses in the ìnstall-dependencies` goal in `pom.xml`, we just set `--ignore-scripts`

### DIFF
--- a/db/package.json
+++ b/db/package.json
@@ -1,7 +1,8 @@
 {
   "name": "deploy",
   "dependencies": {
-    "@sap/hdi-deploy": "^5"
+    "@sap/hdi-deploy": "^5",
+    "@cap-js/hana": "^1"
   },
   "engines": {
     "node": "^18"
--- a/db/package.json
+++ b/db/package.json
@@ -5,7 +5,7 @@
     "@sap/hdi-deploy": "^5"
   },
   "engines": {
-    "node": "^18 || ^20"
+    "node": "^20"
   },
   "scripts": {
     "start": "node node_modules/@sap/hdi-deploy/deploy.js --use-hdb",
--- a/db/package.json
+++ b/db/package.json
@@ -1,7 +1,7 @@
 {
   "name": "deploy",
   "dependencies": {
-    "@sap/hdi-deploy": "4.8.2"
+    "@sap/hdi-deploy": "^5"
   },
   "engines": {
     "node": "^18"
--- a/db/package.json
+++ b/db/package.json
@@ -1,14 +1,14 @@
 {
   "name": "deploy",
   "dependencies": {
-    "@sap/hdi-deploy": "^5",
-    "@cap-js/hana": "^1"
+    "hdb": "^0",
+    "@sap/hdi-deploy": "^5"
   },
   "engines": {
-    "node": "^18"
+    "node": "^18 || ^20"
   },
   "scripts": {
-    "start": "node node_modules/@sap/hdi-deploy/deploy.js",
+    "start": "node node_modules/@sap/hdi-deploy/deploy.js --use-hdb",
     "build": "npm i && npx cds build .. --for hana --production"
   }
 }
--- a/mtx/sidecar/package.json
+++ b/mtx/sidecar/package.json
@@ -1,12 +1,13 @@
 {
   "dependencies": {
     "@sap/cds": "^7",
-    "@sap/cds-mtxs": "^1",
+    "@sap/cds-mtxs": "^2",
     "@sap/xssec": "^3",
     "express": "^4",
     "hdb": "^0",
     "passport": "^0",
-    "@sap/hdi-deploy": "^5"
+    "@sap/hdi-deploy": "^5",
+    "@cap-js/hana": "^1"
   },
   "devDependencies": {
     "sqlite3": "^5"
--- a/mtx/sidecar/package.json
+++ b/mtx/sidecar/package.json
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "@sap/cds": "^7",
+    "@sap/cds": "^8",
     "@sap/cds-mtxs": "^2",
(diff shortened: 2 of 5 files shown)


srv/pom.xml
```diff
@@ -192,7 +192,7 @@
 							<goal>npm</goal>
 						</goals>
 						<configuration>
-							<arguments>install @sap/cds-dk@${cds.install-cdsdk.version} @sap/cds-mtxs@^2 --no-save</arguments>
+							<arguments>install @sap/cds-dk@${cds.install-cdsdk.version} @sap/cds-mtxs@^2 --ignore-scripts --no-save</arguments>
```
mofterdinger, line 195: Is this required to skip the `postInstall` script of the hdi-deploy? I assume it will skip also other scripts which may be required. Do you see any side-effects ?

srv/src/main/resources/application.yaml
```diff
@@ -9,6 +9,7 @@ cds:
   odata-v4:
     endpoint.path: "/api"
   security:
+    authentication.normalize-provider-tenant: true
```
mofterdinger, line 12: Mhh, somehow I don't like this approach to add this configuration to make sed work. Maybe we change our ITest script here instead.

srv/pom.xml
StefanHenke: Yes, exactly. I haven´t seen any side effects of `--ignore-scripts`. The alternative solution would be to install `@cap-js/hana`. This will install the necessary hana db client.

srv/src/main/resources/application.yaml
StefanHenke: I was thinking about this as well, but in a second step. I wanted to at least make the tests green again. Proposal: As we have dedicated tests which also tests tenant isolation ([link](https://github.wdf.sap.corp/cds-java/cds-services/blob/c9169f16569991ea7406c5540ef3d9badc4ddb7e/integration-tests/bookshop-cloud/src/test/java/com/sap/cds/demo/BookshopITest.java#L104)), we create a second subscriber account next to `bookshop-subscriber` and use this one instead of the provider account. This requires changes in several repos because of which I didn´t want to do this before getting tests green again at least once.

srv/pom.xml
mofterdinger: ok, then let's use this approach for now. Thanks for clarification.

srv/src/main/resources/application.yaml
mofterdinger: Ok, I looked into the history of this file and we used already this approach in the past. This config parameter was removed recently. Let's add it again to get the test green. Afterwards we can rework the ITests and remove it again.

mofterdinger (approved):

mtx/sidecar/package.json
```diff
@@ -1,12 +1,13 @@
...
     "@sap/xssec": "^3",
     "express": "^4",
     "hdb": "^0",
     "passport": "^0",
-    "@sap/hdi-deploy": "4.8.2"
+    "@sap/hdi-deploy": "^5",
+    "@cap-js/hana": "^1"
   },
   "devDependencies": {
     "sqlite3": "^5"
```
beckermarc, line 13: Let's use @cap-js/sqlite here as well, this is the dependency section you get with a fresh cds add:

```
  "dependencies": {
    "@cap-js/hana": "^1",
    "@sap/cds": "^8",
    "@sap/cds-mtxs": "^2",
    "@sap/xssec": "^4",
    "express": "^4"
  },
  "devDependencies": {
    "@cap-js/sqlite": "^1"
  },
```

db/package.json
```diff
@@ -1,7 +1,8 @@
 {
```
beckermarc, line 1: Let's use the same that we get generated from a fresh cds init --add java,sample,hana,mta with cds-dk 8:

```json
{
  "name": "deploy",
  "dependencies": {
    "hdb": "^0",
    "@sap/hdi-deploy": "^5"
  },
  "engines": {
    "node": "^18 || ^20"
  },
  "scripts": {
    "start": "node node_modules/@sap/hdi-deploy/deploy.js --use-hdb",
    "build": "npm i && npx cds build .. --for hana --production"
  }
}
```

mtx/sidecar/package.json
beckermarc: Let's also update the Node engines version below to `^18 || ^20`

beckermarc (approved):

db/package.json
```diff
@@ -1,13 +1,14 @@
...
   "name": "deploy",
   "dependencies": {
-    "@sap/hdi-deploy": "4.8.2"
+    "hdb": "^0",
+    "@sap/hdi-deploy": "^5"
   },
   "engines": {
-    "node": "^18"
+    "node": "^18 || ^20"
```
beckermarc, line 8:
```suggestion
    "node": "^20"
```

mtx/sidecar/package.json
```diff
@@ -1,18 +1,16 @@
...
+    "@sap/xssec": "^4",
+    "express": "^4"
   },
   "devDependencies": {
-    "sqlite3": "^5"
+    "@cap-js/sqlite": "^1"
   },
   "engines": {
-    "node": "^18"
+    "node": "^18 || ^20"
```
beckermarc, line 13:
```suggestion
    "node": "^20"
```

PR #346, opened 2024-07-19
Fix dependency versions of npm packages

beckermarc:
- Check in `package-lock.json` files to fix dependencies of Node-based apps.
- Use `moduleLookupDirectories` to find `@sap/cds-mtxs` modules from `mtx/sidecar`
- Use `@sap/cds-dk` with `npx --package` in cds-maven-plugin goals
- Delete `package.json` in root directory.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,7 +27,7 @@
 		<cloud.sdk.version>5.10.0</cloud.sdk.version>
 		<xsuaa.version>3.5.2</xsuaa.version>
 		<cf-java-logging-support.version>3.8.3</cf-java-logging-support.version>
-		<cds.install-cdsdk.version>8.0.2</cds.install-cdsdk.version>
+		<cds.cdsdk-version>8.0.2</cds.cdsdk-version>
 	</properties>
 
 	<modules>
--- a/.cdsrc.json
+++ b/.cdsrc.json
@@ -4,5 +4,8 @@
         "extensibility": true,
         "toggles": true
     },
-    "profile": "with-mtx-sidecar"
+    "profile": "with-mtx-sidecar",
+    "cdsc": {
+        "moduleLookupDirectories": [ "node_modules/", "target/cds/", "mtx/sidecar/node_modules/" ]
+    }
 }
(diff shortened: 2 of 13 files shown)


beckermarc:
ST & MT scenario tested manually in CF.

pom.xml
```diff
@@ -27,7 +27,7 @@
 		<cloud.sdk.version>5.10.0</cloud.sdk.version>
 		<xsuaa.version>3.5.2</xsuaa.version>
 		<cf-java-logging-support.version>3.8.3</cf-java-logging-support.version>
-		<cds.install-cdsdk.version>8.0.2</cds.install-cdsdk.version>
+		<cds.cdsdk-version>8.0.2</cds.cdsdk-version>
```
StefanHenke, line 30: For my understanding as I don´t find usage of this property: is it meant to override the one that is defined in `cds-services`?

StefanHenke (approved):
Just a question for understanding

pom.xml
mofterdinger: Hi, this property is used by the goal `cds`: https://cap.cloud.sap/docs/java/assets/cds-maven-plugin-site/cds-mojo.html#cdsdkVersion
If this property is set. the goal `cds` uses npx to install the package @sap/cds-dk. So we don't need an explicit execution of goal `install-cdsdk`.

mofterdinger:
This new project setup looks good, but somehow it differs more and more from our project template (cds-services-archetype). Would it make sense to adjust also our archetype in that way ?

mofterdinger (approved):

beckermarc:
> This new project setup looks good, but somehow it differs more and more from our project template (cds-services-archetype). Would it make sense to adjust also our archetype in that way ?

This was discussed also [here](https://github.tools.sap/cap/issues/issues/16227). I'd suggest to test drive this new project setup for some time here. If it works nicely we can think about using that in our archetype + cds init/add. However it already has the known limitation of being not really usable for further npm-based plugins at the moment (which we don't use in bookshop).

Issue #354, opened 2024-08-08
Starting JUnit Tests from IntelliJ fails with java.lang.IllegalStateException: java.lang.IllegalStateException: Failed to load ApplicationContext


jschyma:
When I want to execute all the tests using IntelliJ most of them fail due to the following error.
**What I do:**
1. clone the repo
2. open in IntelliJ
3. run mvn install (here everything is ok and works - also the tests)
4. mark the Gen/java folder as generated source location
5. then run IntelliJ Rebuild (works fine)
6. then create a Junit Run Config for all in package my.bookshop (no additional attributes)

**What I would expect:**
If there is something special to be done. I would expect that to be mentioned in the readme. If it is special to IntelliJ to be mentioned in the IntelliJ Section.

**Error:**
[...]

PR #355, opened 2024-08-08
Improve IntelliJ setup instructions

beckermarc:
Closes https://github.com/SAP-samples/cloud-cap-samples-java/issues/354


mofterdinger (approved):
Doesn't IntelliJ have a feature similar to `Maven -> Update project ...' as I know it from Eclipse.

beckermarc:
> Doesn't IntelliJ have a feature similar to `Maven -> Update project ...' as I know it from Eclipse.

It has something like that, but it can't detect the additional source and resource locations that we contribute with our CDS Maven Plugin. That only works in VS Code and Eclipse..

PR #357, opened 2024-08-21
update package-lock.json

StefanHenke:


beckermarc:
Please also update package-lock.json in app folder to solve security alerts around axios.

PR #363, opened 2024-09-11
Use sapmachine for maven build

mofterdinger:
With [latest version](https://github.com/actions/setup-java/releases/tag/v4.3.0) the github action actions/setup-java supports installing sapmachine :-)
I think we should advertise our Java VM in our projects.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -25,6 +25,6 @@ jobs:
       uses: actions/setup-java@v4
       with:
         java-version: 17
-        distribution: 'temurin'
+        distribution: 'sapmachine'
     - name: Build with Maven
       run: mvn -ntp -B clean install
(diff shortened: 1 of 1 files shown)


vmikhailenko (approved):

.github/workflows/maven.yml
```diff
@@ -25,6 +25,6 @@ jobs:
       uses: actions/setup-java@v4
       with:
         java-version: 17
```
beckermarc, line 27: I think we should also in a future PR move to Java 21 in this project

beckermarc (approved):
Nice 👍

.github/workflows/maven.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     - name: Set up JDK
       uses: actions/setup-java@v4
       with:
-        java-version: 17
+        java-version: 21
```
beckermarc, line 27: I think we shouldn't change the version here, without moving the whole project to 21. As said I would tackle that in a different PR

mofterdinger: ah ok, sorry, will revert

PR #377, opened 2024-10-07
Demonstrate cds-feature-attachments with bookshop

mofterdinger:
This PR demonstrates the usage of the cds-feature-attachments in the bookshop.


app/fiori.html
```diff
@@ -14,8 +14,8 @@
 		};
 	</script>
 
-	<script id="sap-ushell-bootstrap" src="https://sapui5.hana.ondemand.com/1.129.0/test-resources/sap/ushell/bootstrap/sandbox.js"></script>
-	<script id="sap-ui-bootstrap" src="https://sapui5.hana.ondemand.com/1.129.0/resources/sap-ui-core.js"
+	<script id="sap-ushell-bootstrap" src="https://sapui5nightly.int.sap.eu2.hana.ondemand.com/test-resources/sap/ushell/bootstrap/sandbox.js"></script>
```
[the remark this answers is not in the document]
mofterdinger: Mhh, not sure, I suggest we wait until SAP UI5 1.131.0 is released. It contains the required fix for the eTag issue.

BraunMatthias (approved):

db/pom.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
beckermarc, line 1: An alternative to this approach would be: Add the covers to the Books entity from a CDS file within the srv directory. I usually would prefer to use the plugins where defined and avoid defining plugin dependencies in root pom.xml. If we discuss that this is not a good pattern (e.g. adding Attachments from a source file within srv directory) we should rather consider changing or making configurable the behaviour of visibility of plugin CDS files (e.g. visible to db folder by default/by configuration in the resolve goal)

beckermarc (changes requested):

mta-multi-tenant.yaml
```diff
@@ -85,6 +85,7 @@ modules:
         url: ~{srv-url}
         forwardAuthToken: true
         strictSSL: true
+        timeout: 600000  # required for long running HTTP requests, e.g. large attachment uploads
```
beckermarc, line 88: 10 minutes might be a little bit too large, increases attack vector for [slow loris attacks](https://www.cloudflare.com/learning/ddos/ddos-attack-tools/slowloris/)

srv/pom.xml
```diff
@@ -74,6 +74,13 @@
 			<artifactId>connectivity-apache-httpclient4</artifactId>
 		</dependency>
 
+		<!-- RE-USE PLUGIN DEPENDENCIES -->
```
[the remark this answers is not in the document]
beckermarc: Yeah, but it belongs to the CDS ecosystem. I don't think we need to distinguish this here. It should be easily discoverable together with the other CDS-related dependencies I think.

mta-multi-tenant.yaml
mofterdinger: what do you suggest ?

srv/pom.xml
```diff
@@ -58,6 +58,12 @@
 			<scope>runtime</scope>
 		</dependency>
 
+		<dependency>
+			<groupId>com.sap.cds</groupId>
+			<artifactId>cds-feature-attachments</artifactId>
+			<version>${cds-feature-attachments.version}</version>
```
beckermarc, line 64:
```suggestion
			<version>${cds-feature-attachments.version}</version>
			<scope>runtime</scope>
```

mta-multi-tenant.yaml
beckermarc: Is the default enough (I suppose 2 minutes) or did you run into a concrete issue?

srv/pom.xml
mofterdinger: done

mta-multi-tenant.yaml
mofterdinger: The default is 30s, which causes issues during the upload of larger attachments. So, what should be the max. upload size we want to support out-of-the-box in the bookshop ?
E.g. I tested with very large attachments (up to 4GB) and it takes several minutes to upload them, depending on the network performance.

beckermarc: 4GB sound very unreasonable to me for a book cover 😆
I guess we should think about showcasing a size restriction (or providing such a validation as part of the attachment plugin out-of-the-box). For smaller-sized images 30s should be enough. I don't think increasing the timeout to 10min should be a good practice we want to showcase here.

mofterdinger: Removed as requested. From my point of view we should demonstrate how to configure this timeout, because I assume stakeholders will run into this issue.

beckermarc (approved):

app/fiori.html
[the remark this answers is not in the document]
mofterdinger: After updating to SAP UI5 1.131.0 the attachments feature works out-of-the-box.

PR #382, opened 2024-10-17
Configuration needed to use IAS and SMS

davidhunglam:
PR only to review the changes, we'll keep the `identity` branch. It's not supposed to be merged to `main`!


davidhunglam:
PR only to review the changes, we'll keep the `identity` branch. It's not supposed to be merged to `main`!

mta-multi-tenant.yaml
```diff
@@ -53,83 +64,107 @@ modules:
...
-      config: # override xsappname as it needs to be unique
-        xsappname: bookshop-mt-${org}-${space}
-        oauth2-configuration:
-          redirect-uris:
-          - https://*.~{app-api/app-domain}/**
-    requires:
-      - name: app-api
-  - name: bookshop-mt-service-manager
+  # -----------------------------------------------------
+  - name: bookshop-service-manager
```
StefanHenke, line 120: from a naming pattern perspective, we chose a name for service instances which makes it easy to correlate to the scenario (e.g. as integration tests for all scenarios are deployed into the same subaccount). What about `bookshop-mt-ias-service-manager`?

https://github.com/SAP-samples/cloud-cap-samples-java/blob/messaging/mta-multi-tenant-messaging.yaml

mta-multi-tenant.yaml
```diff
@@ -53,83 +64,107 @@ modules:
...
-          getDependencies: ~{srv-api/srv-url}/mt/v1.0/subscriptions/dependencies
-          onSubscription: ~{srv-api/srv-url}/mt/v1.0/subscriptions/tenants/{tenantId}
-          onSubscriptionAsync: true
-          onUnSubscriptionAsync: true
-          onUpdateDependenciesAsync: true
-          callbackTimeoutMillis: 3600000
-        displayName: bookshop-java
-        description: A simple CAP Java project.
-        category: 'Category'
+        authorization:
```
StefanHenke, line 131: This is for AMS integration. Currently, we do not have one in the bookshop example. So, in principle it would not be required at the moment. We would require it if we add support for AMS-based authorizations which will happen probably soon.

mta-multi-tenant.yaml
```diff
@@ -53,83 +64,107 @@ modules:
...
+          app-domain: '${default-domain}'
 # --------------------- RESOURCES ---------------------
 resources:
-# -----------------------------------------------------
-  - name: bookshop-mt-uaa
-    type: org.cloudfoundry.managed-service
-    parameters:
-      service: xsuaa
-      service-plan: application
-      path: ./xs-security-mt.json
```
StefanHenke, line 98: I think you could also remove the file `xs-security-mt.json` from the branch.

mta-multi-tenant.yaml
StefanHenke, line 1: In other branches like for messaging, we kept the `mta-multi-tenant.yaml` and added a separate one `mta-multi-tenant-messaging.yaml`. Should we stick to that pattern?

mta-multi-tenant.yaml
BraunMatthias: --> this afternoon ;-)

mta-multi-tenant.yaml
BraunMatthias: Dedicated mta files would be the preferred solution to demonstrate different flavours. I'd vote for `mta-multi-tenant-ias-ams.yaml`

mta-multi-tenant.yaml
davidhunglam: I was wondering whether that's necessary in general, or whether it's possible to share these service instances.

But I guess, regarding the service manager, you'd want a dedicated one. But what about the identity service?

StefanHenke: As you said, it depends on the service type. Service manager might also be reusable, but to avoid interferences between the different test scenarios, I wouldn´t do it. Identity service should not cause issues if we share service instance. Howver, why not just adding something like `...-mt-ias` to all service instances. This would make it easier to manage our integration test subaccount and have consistency with service instance naming.

StefanHenke (dismissed):

PR #389, opened 2024-11-06
Don't hardcode `java` build task

swaldmann:
In the sample here and in `cds add sample` the `srv/pom.xml` build task is hardcoded to `java`.

It would imo be better to simply use `cds build --production`, as the build itself can determine which plugin to run based on the context.

This hardcoded approach doesn't work with build plugins, for example for the AMS integration. The plugin is automatically pulled for `cds build --production` but left out when the `java` build task is explicitly specified.

@mofterdinger @beckermarc

### DIFF
--- a/srv/pom.xml
+++ b/srv/pom.xml
@@ -212,7 +212,7 @@
 						</goals>
 						<configuration>
 							<commands>
-								<command>build --for java</command>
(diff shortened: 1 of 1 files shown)


swaldmann:
Discussing different solution, closing this PR

PR #391, opened 2024-11-13
Update cds-dk, plugins and Postgres

mofterdinger:


mta.yaml
```diff
@@ -82,7 +82,7 @@ resources:
       service: postgresql-db
       service-plan: standard
       config:
-        engine_version: "14"
+        engine_version: "16"
```
vmikhailenko, line 85: We might need to re-create test Postgres instance after that.

vmikhailenko (approved):

mta.yaml
mofterdinger: yes, that could be the case. Last time if I remember correct the instance was updated during deployment. We will see.

mofterdinger: will downgrade to 15 in a new PR.

PR #393, opened 2024-11-20
Demonstrate custom logic for Tree Tables

etimr:
Custom handler for non-HANA DBs.

### DIFF
--- a/srv/src/main/java/my/bookshop/handlers/HierarchyHandler.java
+++ b/srv/src/main/java/my/bookshop/handlers/HierarchyHandler.java
@@ -1,0 +1,236 @@
+package my.bookshop.handlers;
+
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.springframework.context.annotation.Profile;
+import org.springframework.stereotype.Component;
+
+import com.sap.cds.Row;
+import com.sap.cds.ql.CQL;
+import com.sap.cds.ql.Select;
+import com.sap.cds.ql.cqn.CqnElementRef;
+import com.sap.cds.ql.cqn.CqnPredicate;
+import com.sap.cds.ql.cqn.CqnSelect;
+import com.sap.cds.ql.cqn.CqnValue;
+import com.sap.cds.ql.cqn.Modifier;
+import com.sap.cds.ql.cqn.transformation.CqnTopLevelsTransformation;
+import com.sap.cds.ql.cqn.transformation.CqnAncestorsTransformation;
+import com.sap.cds.ql.cqn.transformation.CqnDescendantsTransformation;
+import com.sap.cds.ql.cqn.transformation.CqnFilterTransformation;
+import com.sap.cds.ql.cqn.transformation.CqnSearchTransformation;
+import com.sap.cds.ql.cqn.transformation.CqnOrderByTransformation;
+import com.sap.cds.ql.cqn.transformation.CqnTransformation;
+import com.sap.cds.services.cds.CdsReadEventContext;
+import com.sap.cds.services.cds.CqnService;
+import com.sap.cds.services.handler.EventHandler;
+import com.sap.cds.services.handler.annotations.Before;
+import com.sap.cds.services.handler.annotations.ServiceName;
+import com.sap.cds.services.persistence.PersistenceService;
+
+import cds.gen.adminservice.AdminService_;
+import cds.gen.adminservice.GenreHierarchy;
+import cds.gen.adminservice.GenreHierarchy_;
+
+@Component
+@Profile("default")
+@ServiceName(AdminService_.CDS_NAME)
+public class HierarchyHandler implements EventHandler {
+
+    private final PersistenceService db;
+    private static final Class<GenreHierarchy_> GENRE_HIERARCHY = GenreHierarchy_.class;
+
+    HierarchyHandler(PersistenceService db) {
+		this.db = db;
+	}
+
+    @Before(event = CqnService.EVENT_READ, entity = GenreHierarchy_.CDS_NAME)
+    public void readGenreHierarchy(CdsReadEventContext event) {
+        List<CqnTransformation> trafos = event.getCqn().transformations();
+        List<GenreHierarchy> result = null;
+
+        if (trafos.size() < 1) {
+            return;
+        } else if (getTopLevels(trafos) instanceof CqnTopLevelsTransformation topLevels) {
+            result = topLevels(topLevels, CQL.TRUE);
+        } else if (trafos.get(0) instanceof CqnDescendantsTransformation descendants) {
+            result = handleDescendants(descendants);
+        }  else if (trafos.get(0) instanceof CqnAncestorsTransformation ancestors) {
+            if (trafos.size() == 2 && trafos.get(1) instanceof CqnTopLevelsTransformation topLevels) {
+                result = handleAncestors(ancestors, topLevels);
+            } else if (trafos.size() == 3 && trafos.get(2) instanceof CqnTopLevelsTransformation topLevels) {
+                result = handleAncestors(ancestors, topLevels);
+            }
+        }
+
+        setResult(event, result);
+    }
+
+    private CqnTopLevelsTransformation getTopLevels(List<CqnTransformation> trafos) {
+        if (trafos.get(0) instanceof CqnTopLevelsTransformation topLevels) {
+            return topLevels;
+        } else if (trafos.size() == 2 && trafos.get(0) instanceof CqnOrderByTransformation && trafos.get(1) instanceof CqnTopLevelsTransformation topLevels) {
+            return topLevels;
+        }
+        return null;
+    }
+
+    private void setResult(CdsReadEventContext event, List<GenreHierarchy> result) {
+        if (!result.isEmpty()) {
+            addDrillState(result);
+        }
+
+        event.setResult(result);
+    }
+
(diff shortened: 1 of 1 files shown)


srv/src/main/java/my/bookshop/handlers/HierarchyHandler.java
```diff
@@ -0,0 +1,236 @@
...
+import cds.gen.adminservice.GenreHierarchy;
+import cds.gen.adminservice.GenreHierarchy_;
+
+@Component
+@Profile("default")
+@ServiceName(AdminService_.CDS_NAME)
+public class HierarchyHandler implements EventHandler {
+
+    private final PersistenceService db;
+    private static final Class<GenreHierarchy_> GENRE_HIERARCHY = GenreHierarchy_.class;
```
agoerler, line 46: there should be  `AdminService.GENRE_HIERARCHY`

srv/src/main/java/my/bookshop/handlers/HierarchyHandler.java
```diff
@@ -0,0 +1,236 @@
...
+    @Before(event = CqnService.EVENT_READ, entity = GenreHierarchy_.CDS_NAME)
+    public void readGenreHierarchy(CdsReadEventContext event) {
+        List<CqnTransformation> trafos = event.getCqn().transformations();
+        List<GenreHierarchy> result = null;
+
+        if (trafos.size() < 1) {
+            return;
+        } else if (getTopLevels(trafos) instanceof CqnTopLevelsTransformation topLevels) {
+            result = topLevels(topLevels, CQL.TRUE);
+        } else if (trafos.get(0) instanceof CqnDescendantsTransformation descendants) {
```
agoerler, line 61: maybe

```suggestion
        } 
        
        if (getTopLevels(trafos) instanceof CqnTopLevelsTransformation topLevels) {
            result = topLevels(topLevels, CQL.TRUE);
        } else if (trafos.get(0) instanceof CqnDescendantsTransformation descendants) {
```

srv/src/main/java/my/bookshop/handlers/HierarchyHandler.java
```diff
@@ -0,0 +1,236 @@
...
+        CqnTransformation trafo = ancestors.transformations().get(0);
+        Select<GenreHierarchy_> inner = Select.from(GENRE_HIERARCHY).where(gh -> gh.node_id().eq(CQL.get("$outer.node_id")));
+        if (trafo instanceof CqnFilterTransformation filter) {
+            inner.where(CQL.and(filter.filter(), inner.where().get()));
+        } else if (trafo instanceof CqnSearchTransformation search) {
+            inner.search(search.search());
+        }
+        Select<GenreHierarchy_> outer = Select.from(GENRE_HIERARCHY).columns(gh -> gh.node_id().as("i0"), gh -> gh.parent().node_id().as("i1"),
+            gh -> gh.parent().parent().node_id().as("i2"), gh -> gh.parent().parent().parent().node_id().as("i3"),
+            gh -> gh.parent().parent().parent().parent().node_id().as("i4")).where(gh -> gh.exists($outer -> inner));
```
agoerler, line 140: or using an in-subquery:

```suggestion
        Select<GenreHierarchy_> inner = Select.from(GENRE_HIERARCHY).columns(gh -> gh.node_id());
        if (trafo instanceof CqnFilterTransformation filter) {
            inner.where(filter.filter());
        } else if (trafo instanceof CqnSearchTransformation search) {
            inner.search(search.search());
        }
        Select<GenreHierarchy_> outer = Select.from(GENRE_HIERARCHY).columns(gh -> gh.node_id().as("i0"), gh -> gh.parent().node_id().as("i1"),
            gh -> gh.parent().parent().node_id().as("i2"), gh -> gh.parent().parent().parent().node_id().as("i3"),
            gh -> gh.parent().parent().parent().parent().node_id().as("i4")).where(gh -> gh.node_id().in(inner));
```

srv/src/main/java/my/bookshop/handlers/HierarchyHandler.java
```diff
@@ -0,0 +1,236 @@
...
+        CqnSelect allCqn = Select.from(GENRE_HIERARCHY).where(filter);
+        var all = db.run(allCqn).listOf(GenreHierarchy.class);
+        all.forEach(gh -> lookup.put(gh.getNodeId(), gh));
+        all.forEach(gh -> gh.setParent(lookup.get(gh.getParentId())));
+        all.forEach(gh -> gh.setDistanceFromRoot(dfr(gh)));
+
+        return all.stream().sorted(new Sorter()).toList();
+    }
+
+    private static long dfr(GenreHierarchy gh) {
```
agoerler, line 206:
```suggestion
    private static long distanceFromRoot(GenreHierarchy gh) {
```

srv/src/main/java/my/bookshop/handlers/HierarchyHandler.java
```diff
@@ -0,0 +1,236 @@
...
+        String toPath(GenreHierarchy gh) {
+            String path;
+
+            path = gh.getName();
+            while (gh.getParent() != null) {
+                gh = gh.getParent();
+                path = gh.getName() + "." + path;
+            }
+
+            return path;
```
agoerler, line 232: I think this (i.e. my ☺️) algorithm is buggy, unfortunately. We need to compare level by level

srv/src/main/java/my/bookshop/handlers/HierarchyHandler.java
```diff
@@ -0,0 +1,253 @@
...
+                res = last1.compareTo(last2);
+                if (res != 0) {
+                    return res;
+                }
+            }
+            return res;
+        }
+
+        Stack<String> getPath(GenreHierarchy gh){
+            Stack<String> path = new Stack<>();
```
etimr, line 244: I've tried a `List` here, but taking the last element doesn't look nice. `Deque` looks like an overkill for such use case.

srv/src/main/java/my/bookshop/handlers/HierarchyHandler.java
```diff
@@ -91,7 +93,7 @@ private void setResult(CdsReadEventContext event, List<GenreHierarchy> result) {
     private void addDrillState(List<GenreHierarchy> ghs) {
         List<Integer> ids = ghs.stream().map(gh -> gh.getNodeId()).toList();
         Set<Integer> parents = ghs.stream().map(gh -> gh.getParentId()).filter(p -> p != 0).collect(Collectors.toSet());
-        CqnSelect q = Select.from(GENRE_HIERARCHY, gh -> gh.parent()).columns(gh -> gh.node_id())
+        CqnSelect q = Select.from(AdminService_.GENRE_HIERARCHY, gh -> gh.parent()).columns(gh -> gh.node_id())
```
agoerler, line 96: or use a static import for `AdminService_.GENRE_HIERARCHY`

srv/src/main/java/my/bookshop/handlers/HierarchyHandler.java
agoerler: There is an issue with `java.util.Stack`: `java.util.Stack` extends `java.util.Vector`. But `java.util.Vector` is a very old Java class and is _thread-safe_. It contains a lot of synchronization, which makes it slow. Even worse, synchronization in the Java memory model introduces a memory-barrier for all concurrent threads, which might or might not use `Vector` or `Stack`. And a memory-barrier is something really bad: assume you have concurrent thread, which run in parallel on multiple CPU cores. They typically do their work independent of each other and use their CPU core's registers. If one thread hits a memory barrier all need to flush their registers to memory so that the synchronization can work properly. This is a quite expensive operation.

See: https://stackoverflow.com/questions/12524826/why-should-i-use-deque-over-stack

So, overall: `java.util.Vector` and `java.util.Stack` are evil.

srv/src/main/java/my/bookshop/handlers/HierarchyHandler.java
```diff
@@ -225,13 +226,13 @@ private class Sorter implements Comparator<GenreHierarchy> {
...
-            Stack<String> path2 = getPath(gh2);
+            Deque<String> path1 = getPath(gh1);
+            Deque<String> path2 = getPath(gh2);
             int res = 0;
 
             while (!path1.isEmpty() && !path2.isEmpty()) {
-                String last1 = path1.pop();
-                String last2 = path2.pop();
+                String last1 = path1.removeFirst();
+                String last2 = path2.removeFirst();
```
agoerler, line 235:
```suggestion
                String last1 = path1.pop();
                String last2 = path2.pop();
```

?

srv/src/main/java/my/bookshop/handlers/HierarchyHandler.java
```diff
@@ -240,10 +241,10 @@ public int compare(GenreHierarchy gh1, GenreHierarchy gh2) {
             return res;
         }
 
-        Stack<String> getPath(GenreHierarchy gh){
-            Stack<String> path = new Stack<>();
+        Deque<String> getPath(GenreHierarchy gh){
+            Deque<String> path = new ArrayDeque<>();
             do {
-                path.push(gh.getName());
+                path.addFirst(gh.getName());
```
agoerler, line 247:
```suggestion
                path.push(gh.getName());
```

?

agoerler (commented):
Maybe just stick with the push/pop method names.

srv/src/main/java/my/bookshop/handlers/HierarchyHandler.java
etimr: The changes were made according to the documentation https://docs.oracle.com/javase/8/docs/api/java/util/Deque.html
`push(e) -> addFirst(e)`
`pop() -> removeFirst()`

PR #401, opened 2024-11-26
No projection

agoerler:


app/genres/fiori-service.cds
```diff
@@ -7,8 +7,8 @@ using AdminService from '../../srv/admin-service';
 
 annotate AdminService.GenreHierarchy with @Aggregation.RecursiveHierarchy#GenreHierarchy: {
     $Type: 'Aggregation.RecursiveHierarchyType',
-    NodeProperty: node_id, // identifies a node
-    ParentNavigationProperty: parent // navigates to a node's parent
+    NodeProperty: ID, // identifies a node
+    ParentNavigationProperty: parnt // navigates to a node's parent
```
agoerler, line 11: we have a bug, which prevents us to use `parent` as name for the parent association

db/books.cds
```diff
@@ -47,9 +47,9 @@ annotate Authors with
  */
 entity Genres {
     key ID      : Integer;
-    node        : Integer not null;
-    parent_node : Integer default 0;
     name        : localized String(255);
     descr       : localized String(1000);
-    parent      : Association to one Genres on parent.node = parent_node;
+    parnt    : Association to Genres;
+    children : Composition of many Genres
```
etimr, line 53: Why do we need a composition of many here? At the end we want to have a foreign key in the db table which is a result of `parnt    : Association to Genres;`

agoerler: > Why do we need a composition of many here? At the end we want to have a foreign key in the db table which is the result of `parnt : Association to Genres;`

It's already there [on main](https://github.com/SAP-samples/cloud-cap-samples-java/blob/main/db/books.cds#L52). I kept it to minimize the delta to `main`. But I'm also fine with removing it.

etimr: I've missed that it's already in main. You are right, we don't need to make any additional changes, that are not relevant for us.

etimr (approved):

PR #402, opened 2024-11-26
Consume cds-services 3.5.0

agoerler:


srv/src/main/java/my/bookshop/handlers/AdminServiceHandler.java
```diff
@@ -94,14 +94,6 @@ public void beforeCreateOrder(Stream<Orders> orders, EventContext context) {
 				order.getItems().forEach(orderItem -> {
 					// validation of the Order creation request
 					Integer quantity = orderItem.getQuantity();
-					if (quantity == null || quantity <= 0) {
-						// errors with localized messages from property files
-						// exceptions abort the request and set an error http status code
-						// messages in contrast allow to collect multiple errors
-						messages.error(MessageKeys.QUANTITY_REQUIRE_MINIMUM)
```
beckermarc, line 101: The message removal seems to be missing from this PR.

srv/admin-service.cds
```diff
@@ -32,6 +32,12 @@ annotate AdminService.Books with @cds.search : {
   title
 };
 
+annotate AdminService.OrderItems with  {
+  @mandatory
+  @assert.range: [(0), _]
+  quantity
+};
```
beckermarc, line 39: Should we do this on db-level already? Most other validations like these seem to be defined there.

beckermarc (dismissed):

srv/src/main/java/my/bookshop/handlers/AdminServiceHandler.java
agoerler: The message is still used [here](https://github.com/SAP-samples/cloud-cap-samples-java/pull/402/files?diff=unified&w=0#diff-af608d1373e020896c7b3dfc5904d6e45e8e69219e1efa995396a5f5b060bb25L207). I think this is the warning, which is displayed on the UI. Can we get rid of this as well?

beckermarc: Ah good catch, I guess that can't be easily removed, if we want to keep the warning..

agoerler:
@mofterdinger - please feel invited to review this PR ;-)

mofterdinger (approved):

PR #410, opened 2024-12-10
Downgrade Spring Boot to fix Native Image compilation

beckermarc:


StefanHenke (approved):

pom.xml
```diff
@@ -6,7 +6,7 @@
 	<parent>
 		<groupId>org.springframework.boot</groupId>
 		<artifactId>spring-boot-starter-parent</artifactId>
-		<version>3.3.6</version>
```
mofterdinger, line 9: Would it work with latest version 3.4.0 ?

PR #419, opened 2024-12-19
Restore CodeList

agoerler:


beckermarc:
rebase on main?

db/books.cds
```diff
@@ -47,10 +48,8 @@ annotate Authors with
 /**
  * Hierarchically organized Code List for Genres
  */
-entity Genres {
+entity Genres : sap.common.CodeList {
```
beckermarc, line 51: @OlenaTi: What harm did the value help make, that's implicitly included in CodeList? If it did any harm I think the better approach would be to remove the annotation on `GenreHierarchy` by annotating it with `@cds.odata.valuelist: false`

beckermarc (approved):

agoerler:
Before the merge I need to check if the value-help is still working ...

agoerler:
> Before the merge I need to check if the value-help is still working ...

is working for me.

db/books.cds
etimr: you are right, it also works without `@cds.odata.valuelist: false` now: https://github.com/SAP-samples/cloud-cap-samples-java/pull/420

PR #420, opened 2024-12-19
Restore CodeList aspect

etimr:


beckermarc:
Can we combine all changes in https://github.com/SAP-samples/cloud-cap-samples-java/pull/419?
Is `@cds.odata.valuelist: false` actually required? Did you run into issues without it?

srv/admin-service.cds
```diff
@@ -21,6 +21,7 @@ service AdminService @(requires: 'admin') {
   entity Authors        as projection on my.Authors;
   entity Orders         as select from my.Orders;
   extend my.Genres with Hierarchy;
+  @cds.odata.valuelist: false
```
etimr, line 24: it also works without this annotation now, should I still leave it there to show that we don't use a standard `ValueList` provided by `CodeList` or should I remove it?

beckermarc: If it is not required to be disabled I don't think we need to remove it. Then changes in https://github.com/SAP-samples/cloud-cap-samples-java/pull/419 would be sufficient.

PR #426, opened 2025-01-09
Remove open endpoint /actuator/health

mofterdinger:
By default, CAP Java provides the open endpoints `actuator/health/liveness` and `actuator/health/readiness`. Both can be used by Cf or K8s for application health checks.
There is no need for a public `actuator/health` anymore.

See also in mta.yaml: https://github.com/SAP-samples/cloud-cap-samples-java/blob/228e05f940b4d22acc7d88c862fcb6e8c085ff40/mta-multi-tenant.yaml#L18


beckermarc (approved):
Sounds reasonable. What is missing is showing HOW you would authenticate the /health endpoint correctly, e.g. for intregration into Availability Service. But that must not be part of this PR.

PR #428, opened 2025-01-16
Modification of Hierarchies

etimr:


srv/src/main/java/my/bookshop/handlers/HierarchySiblingActionHandler.java
```diff
@@ -0,0 +1,66 @@
...
+        GenreHierarchy moved = siblingNodes.remove(siblingNodes.indexOf(nodeToMove));
+        nextSibling.ifPresentOrElse(n -> siblingNodes.add(siblingNodes.indexOf(n), moved), () -> siblingNodes.addLast(moved));
+        
+        for (int i=0; i < siblingRanks.size(); i++) {
+            siblingNodes.get(i).setSiblingRank(siblingRanks.get(i));
+        }
+
+        CqnUpdate updateCQN = Update.entity(GenreHierarchy_.class).entries(siblingNodes);
+        db.run(updateCQN);
+        event.setCompleted();
```
agoerler, line 64: I am wondering if we can find a simpler solution but am not sure how. Maybe like this

```suggestion
        CqnStructuredTypeRef ref = event.getCqn().ref; // this identifies the target node
        String nextSiblingId = event.getNextSibling().getId();

        CqnSelect targetNodeWithSiblingsQuery = 
            Select.from(CQL.entity(GenreHierarchy_.class, ref))
                  .items(gh -> gh.ID(), 
                         gh -> gh.parent().children().as("siblings")
                                 .expand(c -> c.ID(), c -> c.siblingRank()).orderBy(c.siblingRank()));
        GenreHierarchy targetNode = db.run(targetNodeWithSiblingsQuery).single(GenreHierarchy.class);
        String targetId = targetNode.getId();
        
        List<GenreHierarchy> newSiblings = new ArrayList(oldSiblings.size());
        for(GenreHierarchy sibling : oldSiblings) {
            if (gh.getId() == targetId) continue;
            if (gh.getId() == nextSiblingId) {
              newSiblings.add(targetNode);
            }
            newSiblings.add(sibling);
        }
        if (targetNode == null) {
              newSiblings.add(targetNode);
        }
        
        int i = 0;
        for(GenreHierarchy sibling : newSiblings) {
            sibling.setSiblingRank(i++);
        }

        CqnUpdate updateCQN = Update.entity(GenreHierarchy_.class).entries(newSiblings);
        db.run(updateCQN);

        event.setCompleted();
```

srv/src/main/java/my/bookshop/handlers/HierarchyExpandHandler.java
```diff
@@ -0,0 +1,53 @@
...
+
+
+
+@Component
+@ServiceName(AdminService_.CDS_NAME)
+/**
+ *  For testing purposes of modifying requests for hierarchies
+ *  Remove when generic solution for $apply and $expand is available
+ */
+@Profile("hybrid")
```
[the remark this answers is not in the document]
etimr: With 3.8.0-m2509 we can already remove this handler

srv/src/main/java/my/bookshop/handlers/HierarchySiblingActionHandler.java
etimr: Row 48 -> we don't have `.children` in `GenreHierarchy` (the composition is excluded), I've tried to change the whole handler to db entity `Genres`, but I'm not sure that this is a better solution.  

Rows 43, 55 -> there is a relatively high probability, that `nextSiblingId` is `null` (every time the node is moved to the end of the siblings list), which is handled in the first solution. 

Rows 67, 68 -> we didn't want to assume, that `siblingRanks` have a step equal to 1, I think overwriting them with existing values is better than generate new ones.

srv/src/main/java/my/bookshop/handlers/HierarchyExpandHandler.java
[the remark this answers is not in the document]
agoerler: go ahead!

eugene-andreev:
Some suggestions https://github.com/SAP-samples/cloud-cap-samples-java/pull/446

app/admin/fiori-service.cds
```diff
@@ -87,6 +96,12 @@ annotate AdminService.GenreHierarchy with {
   ID @UI.Hidden;
 };
 
+@Hierarchy.RecursiveHierarchyActions #GenreHierarchy: {
+  $Type                  : 'Hierarchy.RecursiveHierarchyActionsType',
+  // any name can be the action name with namespace/no bound action name
+  ChangeNextSiblingAction: 'AdminService.moveSiblingAction',
```
beckermarc, line 102: Maybe rename to `changeNextSibling`. Just a minor thing, but having "Action" in an "Action" name sounds a little weird to me.

app/admin/fiori-service.cds
```diff
@@ -96,7 +111,46 @@ annotate AdminService.GenreHierarchy with @UI: {
     LineItem               : [{
         $Type: 'UI.DataField',
         Value: name,
-    }]
+        Label : 'Genre'
```
beckermarc, line 114: Use an i18n key here.

app/admin/fiori-service.cds
```diff
@@ -96,7 +111,46 @@ annotate AdminService.GenreHierarchy with @UI: {
...
+    }],
+};
+
+annotate AdminService.ContentsHierarchy with @UI: {
+    PresentationVariant  : {
+        $Type         : 'UI.PresentationVariantType',
+        RequestAtLeast: [name],
+        Visualizations: ['@UI.LineItem', ],
```
beckermarc, line 122:
```suggestion
        Visualizations: ['@UI.LineItem'],
```

app/admin/fiori-service.cds
```diff
@@ -96,7 +111,46 @@ annotate AdminService.GenreHierarchy with @UI: {
...
+        Value: name,
+        },
+        {
+      $Type: 'UI.DataField',
+      Value: page,
+    }],
+    HeaderInfo            : {
+        $Type         : 'UI.HeaderInfoType',
+        TypeName      : 'Contents Level',
+        TypeNamePlural: 'Contents Levels',
```
beckermarc, line 135: Use i18n keys here and also in the rest of this annotation to enable translation at least into german.

app/admin/webapp/manifest.json
```diff
@@ -119,11 +119,34 @@
                                     "detail" : {
                                         "route" : "AuthorsDetails"
```
beckermarc, line 120: I think this route and AuthorsDetails can be removed.

app/browse-genres/webapp/manifest.json
```diff
@@ -0,0 +1,129 @@
...
+				"settings": {
+					"odataVersion": "4.0"
+				}
+			}
+		},
+		"-sourceTemplate": {
+			"id": "ui5template.basicSAPUI5ApplicationProject",
+			"-id": "ui5template.smartTemplate",
+			"-version": "1.40.12"
+		},
```
beckermarc, line 24:
```suggestion
```

app/browse-genres/webapp/manifest.json
```diff
@@ -0,0 +1,129 @@
...
+                                "@com.sap.vocabularies.UI.v1.LineItem": {
+                                    "tableSettings": {
+                                        "hierarchyQualifier": "GenreHierarchy",
+                                        "type": "TreeTable"
+                                    }
+                                }
+                            }
+						}
+					}
+				},
```
beckermarc, line 103: weird indentation here

app/browse-genres/webapp/manifest.json
```diff
@@ -0,0 +1,129 @@
...
+				}
+			],
+			"targets": {
+				"GenreHierarchyList": {
+					"type": "Component",
+					"id": "GenreHierarchyList",
+					"name": "sap.fe.templates.ListReport",
+					"options": {
+						"settings" : {
+							"entitySet" : "GenreHierarchy",
```
beckermarc, line 85:
```suggestion
							"contextPath" : "/GenreHierarchy",
```

app/browse-genres/webapp/manifest.json
```diff
@@ -0,0 +1,129 @@
...
+						}
+					}
+				},
+				"GenreHierarchyDetails": {
+					"type": "Component",
+					"id": "GenreHierarchyDetails",
+					"name": "sap.fe.templates.ObjectPage",
+					"options": {
+						"settings" : {
+							"entitySet": "GenreHierarchy"
```
beckermarc, line 110:
```suggestion
							"contextPath": "/GenreHierarchy"
```

app/browse/fiori-service.cds
```diff
@@ -161,6 +161,19 @@ annotate CatalogService.Reviews with @(UI : {
...
+annotate CatalogService.GenreHierarchy with @UI: {
+    PresentationVariant  : {
+        $Type         : 'UI.PresentationVariantType',
+        RequestAtLeast: [name],
+        Visualizations: ['@UI.LineItem', ],
+    },
+    LineItem               : [{
+        $Type: 'UI.DataField',
+        Value: name,
+        Label : 'Genre'
```
beckermarc, line 173: i18n

app/browse/fiori-service.cds
```diff
@@ -161,6 +161,19 @@ annotate CatalogService.Reviews with @(UI : {
     ]}
 });
 
+annotate CatalogService.GenreHierarchy with @UI: {
+    PresentationVariant  : {
+        $Type         : 'UI.PresentationVariantType',
+        RequestAtLeast: [name],
+        Visualizations: ['@UI.LineItem', ],
```
beckermarc, line 168:
```suggestion
        Visualizations: ['@UI.LineItem'],
```

db/data/my.bookshop-Genres.csv
```diff
@@ -1,31 +1,34 @@
...
-120;100;Contemporary Fiction
-121;100;Magical Realism
-122;100;Literary Fiction
-128;100;Fairy Tale
-200;;Non-Fiction
-201;200;Biography
-202;201;Autobiography
-203;200;Essay
-204;200;Speech
+ID;parent_ID;name;siblingRank
```
beckermarc, line 1: Its always a little confusing when debugging, when so many UUIDs look so similar. Maybe actually generate random UUIDs for all of these genres.

srv/src/main/java/my/bookshop/handlers/HierarchySiblingActionHandler.java
```diff
@@ -0,0 +1,75 @@
...
+                .columns(c -> c.ID(), c -> c.siblingRank())
+                .where(c -> c.parent_ID().eq(parentId)))
+                .listOf(GenreHierarchy.class);
+
+        // Shift the sibling ranks to reflect the sort order
+        List<Integer> siblingRanksCopy = siblingNodes.stream().map(ch -> ch.getSiblingRank()).toList();
+
+        String nextSiblingId = event.getNextSibling() == null ? null : event.getNextSibling().getId();
+        Optional<GenreHierarchy> nextSibling = siblingNodes.stream().filter(el -> el.getId().equals(nextSiblingId))
+                .findFirst();
```
beckermarc, line 59:
```suggestion
        Optional<GenreHierarchy> nextSibling = event.getNextSibling() == null ? Optional.empty() : siblingNodes.stream().filter(el -> el.getId().equals(event.getNextSibling().getId()))
                .findFirst();
```

beckermarc: You could also iterate over siblingNodes and instead of finding the `nextSibling` find directly the index of `nextSibling`. As that is the only information you need further down.

srv/src/main/java/my/bookshop/handlers/HierarchySiblingActionHandler.java
```diff
@@ -0,0 +1,75 @@
...
+
+        GenreHierarchy nodeToMove = siblingNodes.stream().filter(el -> idToMove.equals(el.getId())).findFirst().get();
```
beckermarc, line 61: Also here, instead of finding the node, and then using the instance to determine its index, iterate over the `siblingNodes` once and directly figure out the index of the nodeToMove, as this is again the only information you need further down.

srv/src/main/java/my/bookshop/handlers/HierarchySiblingActionHandler.java
```diff
@@ -0,0 +1,75 @@
...
+        GenreHierarchy moved = siblingNodes.remove(siblingNodes.indexOf(nodeToMove));
+        // Exchange siblings
+        nextSibling.ifPresentOrElse(n -> siblingNodes.add(siblingNodes.indexOf(n), moved),
+                () -> siblingNodes.addLast(moved));
+        // Apply ranks
+        for (int i = 0; i < siblingRanksCopy.size(); i++) {
+            siblingNodes.get(i).setSiblingRank(siblingRanksCopy.get(i));
+        }
```
beckermarc, line 69: Why not number children from 1 to n?

app/admin/fiori-service.cds
beckermarc: Ah I meant to rename `moveSiblingAction` to `changeNextSibling`, but `moveSibling` is also fine with me :)

srv/src/main/java/my/bookshop/handlers/HierarchySiblingActionHandler.java
etimr: @beckermarc then it would be something like this for `nextSibling`:
```
   int index = siblingNodes.size() -1;
        if (event.getNextSibling() != null) {
                for (int i = 0; i < siblingNodes.size(); i++) {
                        if ( siblingNodes.get(i).getId().equals(event.getNextSibling().getId())) {
                          index = i;
                          break;
                        }
                }
        }
...
        siblingNodes.add(index, moved);
```
I can't find `nodeToMove` in the same loop because of the `event.getNextSibling() != null` condition. I'm not sure that two loops are better than the previous solution. What do you think?

srv/src/main/java/my/bookshop/handlers/HierarchySiblingActionHandler.java
etimr: @beckermarc I don't get the question :( `siblingRanksCopy` have the same size as the `siblingNodes`. 

If we don't want to iterate through all children, we need several `if` statements: if we are moving an element forward, we need to recalculate all ranks starting from the index of the `siblingRank`, otherwise starting from the index of the `nodeToMove`. 

On our small data set it's easier to recalculate all ranks.

etimr: @beckermarc do you mean remove `siblingRanksCopy` and overwrite existing ranks?
```
        for (int i = 0; i < siblingNodes.size(); i++) {
            siblingNodes.get(i).setSiblingRank(i);
        }
```
It looks easier and one raw of code less, but Evgeny and I wanted to show how existing `siblingRanks` can be reused.

beckermarc: What should it mean to reuse existing `siblingRanks`? In this specific example is there any meaning attached to these ranks? If not I think we should try to keep the code as simple as possible. BTW: Where are these ranks set in case a new node is created?

app/admin/fiori-service.cds
```diff
@@ -138,6 +196,26 @@ annotate AdminService.Books.texts {
...
+//
+//  Annotations for hierarchy ContentsHierarchy
+//
+annotate AdminService.ContentsHierarchy with @Aggregation.RecursiveHierarchy#ContentsHierarchy: {
+    $Type: 'Aggregation.RecursiveHierarchyType',
+    NodeProperty: ID, // identifies a node
+    ParentNavigationProperty: parent // navigates to a node's parent
+  };
+
+  annotate AdminService.ContentsHierarchy with @Hierarchy.RecursiveHierarchy#ContentsHierarchy: {
```
beckermarc, line 209:
```suggestion
annotate AdminService.ContentsHierarchy with @Hierarchy.RecursiveHierarchy#ContentsHierarchy: {
```

app/browse-genres/fiori-service.cds
```diff
@@ -0,0 +1,22 @@
...
+using CatalogService from '../../srv/cat-service';
+
+
+annotate CatalogService.GenreHierarchy with @Aggregation.RecursiveHierarchy#GenreHierarchy: {
+    $Type: 'Aggregation.RecursiveHierarchyType',
+    NodeProperty: ID, // identifies a node
+    ParentNavigationProperty: parent // navigates to a node's parent
+  };
+
+  annotate CatalogService.GenreHierarchy with @Hierarchy.RecursiveHierarchy#GenreHierarchy: {
```
beckermarc, line 14:
```suggestion
annotate CatalogService.GenreHierarchy with @Hierarchy.RecursiveHierarchy#GenreHierarchy: {
```

srv/src/main/java/my/bookshop/handlers/HierarchySiblingActionHandler.java
beckermarc: I made a suggestion which moves both into the same loop: https://github.com/SAP-samples/cloud-cap-samples-java/pull/449

db/data/my.bookshop-Contents.csv
```diff
@@ -0,0 +1,64 @@
+ID;parent_ID;name;page;book_ID
+f846b0b9-01d4-4f6d-82a4-d79204f62514;;Foreword;3;9b084139-0b1e-43b6-b12a-7b3669d75f02
```
beckermarc, line 2: Also here would be nice, if UUIDs would not look so similar everywhere, but that is optional.

beckermarc (approved):

PR #429, opened 2025-01-17
Refactor destination for mocked remote service

StefanHenke:
https://github.wdf.sap.corp/cds-java/home/issues/1887

### DIFF
--- a/srv/src/main/java/my/bookshop/config/DestinationConfiguration.java
+++ b/srv/src/main/java/my/bookshop/config/DestinationConfiguration.java
@@ -1,7 +1,16 @@
 package my.bookshop.config;
 
-import java.util.Map;
-
+import com.sap.cds.services.runtime.CdsRuntime;
+import com.sap.cloud.environment.servicebinding.api.ServiceBinding;
+import com.sap.cloud.sdk.cloudplatform.connectivity.BtpServiceOptions;
+import com.sap.cloud.sdk.cloudplatform.connectivity.DefaultDestinationLoader;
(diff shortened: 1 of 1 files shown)


t-bonk (approved):
LGTM

srv/src/main/java/my/bookshop/config/DestinationConfiguration.java
```diff
@@ -36,44 +36,37 @@ public class DestinationConfiguration {
...
+		String destinationName = environment
+				.getProperty("cds.remote.services.'[API_BUSINESS_PARTNER]'.destination.name");
 
 		logger.info("Destination name for mocked API_BUSINESS_PARTNER: {}", destinationName);
+		logger.info("UAA Service Binding: {} / {}", xsuaaBinding.getName().get(), xsuaaBinding.getServiceName().get());
 
-		ServiceBinding uaaBinding = runtime.getEnvironment().getServiceBindings().filter(b -> b.getServiceName().get().equals("xsuaa")).findFirst().get();
-		logger.info("UAA Service Binding: {} / {}", uaaBinding.getName().get(), uaaBinding.getServiceName().get());
+		ServiceBindingDestinationOptions options = ServiceBindingDestinationOptions.forService(xsuaaBinding)
+				.withOption(BtpServiceOptions.AuthenticationServiceOptions.withTargetUri("http://localhost"))
```
beckermarc, line 53: To be on the safe-side I think you should also add the port to the URL here, retrieved as for `registerLocalDestination()`. Not sure if this even works right now, as I think CF uses port 8080 by default and not 80.

PR #439, opened 2025-02-11
Showase type cds.Map

mofterdinger:
- Enhanced Books and Authors with details map
- Provide initial data for the details fields in csv files


srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -187,7 +213,8 @@ private void discountBooksWithMoreThan111Stock(Books b, boolean premium) {
 
 	private void loadStockIfNotSet(Books b) {
 		if (b.getId() != null && b.getStock() == null) {
-			b.setStock(db.run(Select.from(BOOKS).byId(b.getId()).columns(Books_::stock)).single(Books.class).getStock());
+			b.setStock(
+					db.run(Select.from(BOOKS).byId(b.getId()).columns(Books_::stock)).single(Books.class).getStock());
```
agoerler, line 216: We could just add "stock" also [here](https://github.com/SAP-samples/cloud-cap-samples-java/pull/439/commits/47fbbbdd3a2cb670adaf78b288bee87d6ef613d2#diff-a1d8a452c4e97ab688ec6b9ad6cd40c2c5129a4f73ba2a1b0356e98904555a13R86)

agoerler (approved):

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -127,11 +143,20 @@ public void afterAddReview(BooksAddReviewContext context) {
...
 		books.filter(b -> b.getTitle() != null).forEach(b -> {
-			loadStockIfNotSet(b);
 			discountBooksWithMoreThan111Stock(b, featureToggles.isEnabled("discount"));
 		});
 	}
 
+	@After(event = CqnService.EVENT_READ)
+	public void bigBooks(Stream<Books> books) {
+		books.filter(b -> b.getDetails() != null).forEach(b -> {
+			int pages = (int) b.getDetails().get("pages");
```
beckermarc, line 153: What if pages are not set, but `null`? This will crash the request. There is no guarantee that pages are set, especially for new books.

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
```diff
@@ -74,6 +77,19 @@ class CatalogServiceHandler implements EventHandler {
 		this.analyzer = CqnAnalyzer.create(model);
 	}
 
+	@Before(entity = Books_.CDS_NAME)
+	public void beforeReadBooks(CdsReadEventContext context) {
+		CqnSelect copy = CQL.copy(context.getCqn(), new Modifier() {
+			@Override
+			public List<CqnSelectListItem> items(List<CqnSelectListItem> items) {
+				items.add(CQL.get("details"));
+				items.add(CQL.get("stock"));
```
beckermarc, line 86: What if those are already selected? I think this logic simply puts this onto the select list twice or is there some automatic deduplication?

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
mofterdinger: I tried it with a new book and it worked, because details is already null and the filter catches this already. With the current bookshop I don't think it breaks here. Ok, we can add an additional null check for pages.

beckermarc: I think I find this a strange concept here in general. Isn't the idea behind `cds.Map` that the data is generic and arbitrary and thus I can't really write business logic on it? From a modelling perspective I'd say a better practice would be to model `pages` explicitly. I get that this is a showcase of the feature, but I somehow fear that this example might lead to inappropriate usage. @agoerler what do you think?

srv/src/main/java/my/bookshop/handlers/CatalogServiceHandler.java
mofterdinger: Good question, I tried it with List.contains, but this doesn't seem to work. It adds the same item twice. Looks like there is no sufficient implementation of equals for the CqnSelectListItem implementations.
Not sure what is the best practice here.

mofterdinger: ah, forget my last comment, it works with List.contains. Stupid error on my side using VSCode without auto.build enabled :-).

PR #443, opened 2025-02-28
Leftovers from demonstrating cds.Map

mofterdinger:
- Remove field details from Books entity
- Avoid duplicate items in select list


mofterdinger:
This PR provides just some fixes for review feedback, that was mentioned after the feature was already merged. If we decide to provide another showcase for the cds.Map, I suggest to do it in a different PR.

agoerler:
It just crossed by mind that @OlenaTi is also introducing "pages" in this PR: https://github.com/SAP-samples/cloud-cap-samples-java/pull/428/files#diff-d92b8eb29957a03ad9d23b01feff79da293e30183d0b8eac5d893d95632e30dfR62

Maybe we should store something else in the map then.

etimr:
> It just crossed by mind that @OlenaTi is also introducing "pages" in this PR: https://github.com/SAP-samples/cloud-cap-samples-java/pull/428/files#diff-d92b8eb29957a03ad9d23b01feff79da293e30183d0b8eac5d893d95632e30dfR62
> 
> Maybe we should store something else in the map then.

I'm not sure that there is a conflict between these two things. I'm introducing a page number in the table of context, here, if I get it correctly, we are talking about the page count in the whole book.
Screenshot 2025-02-28 at 14 48 51

beckermarc (approved):
Looks good to me, but it doesn't keep much of Map left (which is fine with me).

PR #445, opened 2025-03-05
Update CDS Services to 3.8.0 and enable Draft Messages

beckermarc:
- Updates @sap/cds-dk to 8.8.0 and CDS Services to 3.8.0
- Enables Draft Messages
- Enables OData containment

### DIFF
--- a/app/addresses/webapp/manifest.json
+++ b/app/addresses/webapp/manifest.json
@@ -86,7 +86,7 @@
                     "name": "sap.fe.templates.ListReport",
                     "options": {
                         "settings" : {
-                            "entitySet" : "Addresses",
+                            "contextPath" : "/Addresses",
                             "navigation" : {
                                 "Addresses" : {
                                     "detail" : {
@@ -103,7 +103,7 @@
                     "name": "sap.fe.templates.ObjectPage",
                     "options": {
                         "settings" : {
-                            "entitySet" : "Addresses",
+                            "contextPath" : "/Addresses",
                             "navigation" : {
                                 "notes" : {
                                     "detail" : {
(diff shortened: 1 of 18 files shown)


app/addresses/webapp/manifest.json
```diff
@@ -86,7 +86,7 @@
                     "name": "sap.fe.templates.ListReport",
                     "options": {
                         "settings" : {
-                            "entitySet" : "Addresses",
+                            "contextPath" : "/Addresses",
```
mofterdinger, line 89: Is this change required due to the enabled containment in OData ?

srv/notes-mashup.cds
```diff
@@ -14,7 +14,7 @@ entity my.bookshop.NoteableAddresses as select from API_BUSINESS_PARTNER.A_Busin
   @readonly PostalCode as postalCode,
   @readonly StreetName as street,
   @readonly HouseNumber as houseNumber,
-  notes
+  notes @(odata.contained: false)
```
mofterdinger, line 17: This disables containment for notes entity, right ? Why is this required ?

app/addresses/webapp/manifest.json
beckermarc: Yes, it is also the recommended property by Fiori Elements

srv/notes-mashup.cds
beckermarc: Yes, because we want to have a UI that uses the list of Notes as the entry point. If containment of notes is enabled (as it is a composition of Addresses) accessing the list of all notes is no longer possible.

mofterdinger (approved):
Nice improvement of the bookshop :-)

PR #451, opened 2025-03-11
Allow to specify a dedicated version for cds-feature-event-hub

StefanHenke:
In order to use the OSS variant of `cds-feature-event-hub` it needs to be possible to specify its version independently from the regular cds version

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,6 +24,7 @@
 		<!-- DEPENDENCIES VERSION -->
 		<jdk.version>21</jdk.version>
 		<cds.services.version>3.6.0</cds.services.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -24,6 +24,7 @@
 		<!-- DEPENDENCIES VERSION -->
 		<jdk.version>21</jdk.version>
 		<cds.services.version>3.6.0</cds.services.version>
+		<cds-feature-event-hub.version>3.6.0</cds-feature-event-hub.version>
```
mofterdinger, line 27:
```suggestion
		<cds-feature-event-hub.version>${cds.services.version}</cds-feature-event-hub.version>
```
As long it's in sync with cds-services, use this approach.

Issue #452, opened 2025-03-12
Upload of attachments fails with error: Provide the missing value.


mofterdinger:
Hi,

with latest version of the Bookshop the upload of Attachments is failing with an error:

<img width="327" alt="Image" src="https://github.com/user-attachments/assets/16e5772a-3129-4685-b245-42a5117fd934" />

In the logs I found this error:

```
Caused by: com.sap.cds.services.utils.ErrorStatusException: Value of element 'up__ID' in entity 'AdminService.Books.covers_drafts' is required
        at com.sap.cds.services.impl.utils.ValidatorErrorUtils.handleValidationError(ValidatorErrorUtils.java:34)
        at com.sap.cds.services.impl.utils.NotNullValidator.handleNotNullError(NotNullValidator.java:163)
        at com.sap.cds.services.impl.utils.NotNullValidator.lambda$assertNotNull$3(NotNullValidator.java:122)
        at com.sap.cds.impl.DataProcessor$5.validate(DataProcessor.java:178)
        at com.sap.cds.impl.DataProcessor$5.lambda$entries$0(DataProcessor.java:169)
        at com.sap.cds.impl.DataProcessor.lambda$validationHandler$0(DataProcessor.java:202)
        at com.sap.cds.impl.DataProcessor$5.lambda$entries$1(DataProcessor.java:169)
        at com.sap.cds.impl.DataProcessor.lambda$forAllEntries$5(DataProcessor.java:238)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
        at com.sap.cds.impl.DataProcessor.lambda$forAllEntries$6(DataProcessor.java:237)
        at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
        at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179)
```

The whole stackrace:

<details>

```
2025-03-12T13:25:19.924+01:00 DEBUG 50491 --- [nio-8080-exec-8] c.s.c.a.o.v4.processors.CdsProcessor     : Value of element 'up__ID' in entity 'AdminService.Books.covers_drafts' is required (service 'PersistenceService$Default', event 'UPDATE', entity 'AdminService.Books.covers_drafts')

[...]

PR #455, opened 2025-03-12
Restore up__ID for covers

beckermarc:
Fixes https://github.com/SAP-samples/cloud-cap-samples-java/issues/452 

Workaround, that is currently required to restore the up__ID, which is no longer provided through the OData API due to containment.


mofterdinger (approved):
Thanks, it solves the upload issue with the cds-feature-attachments.

PR #456, opened 2025-03-14
WIP: dynamic UI.Hidden annotation

etimr:


srv/src/main/java/my/bookshop/handlers/SetHiddenHandler.java
```diff
@@ -0,0 +1,38 @@
...
+
+import cds.gen.adminservice.AdminService_;
+import cds.gen.adminservice.Books_;
+
+@Component
+@ServiceName(AdminService_.CDS_NAME)
+/**
+ * Example of a custom handler for nextSiblingAction
+ */
+@Profile("hybrid")
```
beckermarc, line 24: I think in other areas we are using `cloud`

app/admin/fiori-service.cds
```diff
@@ -39,10 +39,11 @@ annotate AdminService.Books with @(UI : {
         },
         {   
             // TODO: should work dynamically
-            @UI.Hidden,
             $Type  : 'UI.ReferenceFacet',
             Label  : '{i18n>Contents}',
-            Target : 'contents/@UI.PresentationVariant'
+            Target : 'contents/@UI.PresentationVariant',
+            ![@UI.Hidden] : contentsHidden
```
beckermarc, line 45: I'm wondering if we can also refer to a `Singleton` entity here.

srv/src/main/java/my/bookshop/handlers/SetHiddenHandler.java
```diff
@@ -0,0 +1,38 @@
...
+public class SetHiddenHandler implements EventHandler {
+
+    @After(entity = Books_.CDS_NAME)   
+    void removeUIHidden(EventContext context){
```
beckermarc, line 28: I think you should be able to use `Result` in the method signature.

beckermarc:
Replaced by https://github.com/SAP-samples/cloud-cap-samples-java/pull/460 which however still runs into a UI bug. Did you observe the same UI issues with this approach?

PR #460, opened 2025-03-26
Use singleton to show tree table only on HANA

beckermarc:
Replaces https://github.com/SAP-samples/cloud-cap-samples-java/pull/456 by using a singleton, instead of adding a property to books.

### DIFF
--- a/srv/src/main/java/my/bookshop/handlers/AdminServiceHandler.java
+++ b/srv/src/main/java/my/bookshop/handlers/AdminServiceHandler.java
@@ -16,6 +16,7 @@ import java.util.UUID;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
 
+import org.springframework.core.env.Environment;
 import org.springframework.stereotype.Component;
 
 import com.sap.cds.Result;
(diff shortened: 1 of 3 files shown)


beckermarc:
@etimr Can you have a look please. I successfully tested this, now that I avoided the $Not.

srv/src/main/java/my/bookshop/handlers/AdminServiceHandler.java
```diff
@@ -301,6 +304,13 @@ public void restoreCoversUpId(CqnStructuredTypeRef ref, BooksCovers cover) {
 		cover.setUpId((String) analyzer.analyze(ref).rootKeys().get(Books.ID));
 	}
 
+	@On(event = CqnService.EVENT_READ, entity = Info_.CDS_NAME)
```
mofterdinger, line 307: Could be shortened to:
```suggestion
	@On(entity = Info_.CDS_NAME)
```

beckermarc: Yeah, or we remove `CdsReadEventContext` from the signature, as it is not used.

mofterdinger: yes, also fine with me. I didn't recognize the auto merge, sorry.

PR #463, opened 2025-04-03
introduce a dependabot file

rjayasinghe:


beckermarc (approved):

.github/dependabot.yml
```diff
@@ -5,31 +5,25 @@
...
-    groups:
-      dev-dependencies:
-        patterns: 
-          - "*"
-  - package-ecosystem: "npm"
-    directory: "/app"
+  - package-ecosystem: "maven"
+    directories: 
+    - "/"
+    - "/srv"
```
mofterdinger, line 11: I tried also this approach in cds-feature-attachments with multiples directories and it ended up in duplicate PR for the same dependencies, if there is a new version. We will see if it causes the same experience in the bookshop.
I think it was caused by the parent relationship between srv and root modules.

mofterdinger: Yes, I saw the PRs and they are looking good. I will try it again in cds-feature-attachments with the depandabot config from this bookshop.

PR #468, opened 2025-04-09
Ignore updates to express 5.x

mofterdinger:
Taken from this example: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/controlling-dependencies-updated#specifying-the-semantic-versioning-level-to-ignore

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -29,7 +29,7 @@ updates:
         - "*"
     ignore:
       - dependency-name: "express"
-        # For Express, ignore all updates for version 5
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -27,6 +27,10 @@ updates:
       all-npm-dependencies:
         patterns: 
         - "*"
+    ignore:
+      - dependency-name: "express"
+        # For Express, ignore all updates for version 5
```
StefanHenke, line 32:
```suggestion
        # For Express, ignore all updates for version 5 becaue @sap/cds requires express version 4
```

StefanHenke (approved):

PR #492, opened 2025-05-27
Fix hierarchies for draft-enabled

eugene-andreev:
Fixes: <stackenterprise> `questions/73006/73015`
Fallback to plain table in case of draft requests


beckermarc (approved):

kavya-b:
Hi @eugene-andreev I pulled in the changes from this code, and I now see a flat list of the genres: 
image
We were seeing a hierarchical list previously(ignore the second column which has my local changes from few months back; this is an older screen grab):
hierarchy-sample

PR #484, opened 2025-05-07
Prepare new MTX sidecar defaults for CDS 9

swaldmann:
– Do not merge until CDS 9 is out –

CDS 9 will allow us to simplify the sidecar configuration a little bit, as `requires.auth = "dummy"` for `development` in Java projects by default.

PR #498, opened 2025-06-06
Use CAP Java 4.0.2

mofterdinger:


beckermarc:
Let's update consistently to cds9 CAP Java 4 in a single PR.

mofterdinger:
> Let's update consistently to cds9 CAP Java 4 in a single PR. And let's also make sure we remove properties that are now set to the default.

yes, this PR is not finished yet ;-)

beckermarc:
Ah sorry, I always get notifications for all PRs created in this repository. I thought you already added me as reviewer ;)

beckermarc:
Just as a hint: You can also take changes in https://github.com/SAP-samples/cloud-cap-samples-java/pull/484 into account and merge that into this one:

package.json
```diff
@@ -3,7 +3,8 @@
 	"version": "1.0.0",
 	"description": "Generated by cds-services-archetype",
 	"devDependencies": {
-		"@sap/cds-dk": "^8"
+		"@sap/cds-dk": "^9",
+		"@sap/cds-mtxs": "^3"
 	},
 	"workspaces": [
```
mofterdinger, line 9: @beckermarc Shall I remove the workspace or keep ?

beckermarc: We should keep it, as this is still the recommended setup and the `@sap/cds-mtxs` dependency here will hopefully vanish again. If we know that it will not vanish, we can remove the workspace later.

.cdsrc.json
beckermarc, line 1: newParser is now also the default

beckermarc (commented):
I think we should also check the `application.yaml`. There are likely some properties, which are no longer required, as they are defaults now.

package.json
mofterdinger: ok, will keep it.

.cdsrc.json
mofterdinger: thanks, is removed from cdsrc.json

mofterdinger:
> I think we should also check the `application.yaml`. There are likely some properties, which are no longer required, as they are defaults now.

I did a look into the `application.yaml` and in our migration guide 3 -> 4 and it looks like, no changes are required in application.yaml

PR #601, opened 2026-01-27
Use @hierarchy annotation

agoerler:


eugene-andreev:
There is still an issue in 4.8.1 with the error: `Caused by: org.h2.jdbc.JdbcSQLSyntaxErrorException: Column "MY.SIBLINGRANK" not found`
Tested with 4.9.x - works

eugene-andreev (approved):

app/admin/fiori-service.cds
```diff
@@ -94,7 +94,7 @@ annotate AdminService.GenreHierarchy with {
   ID @UI.Hidden;
 };
 
-annotate AdminService.GenreHierarchy with @Hierarchy.RecursiveHierarchyActions #GenreHierarchy: {
+annotate AdminService.GenreHierarchy with @Hierarchy.RecursiveHierarchyActions #GenreHierarchyHierarchy: {
```
agoerler, line 97: I think we should avoid the duplicate HierarchyHierarchy

app/admin/fiori-service.cds
```diff
@@ -104,7 +104,7 @@ annotate AdminService.GenreHierarchy with @UI: {
     PresentationVariant #VH: {
         $Type                      : 'UI.PresentationVariantType',
         Visualizations             : ['@UI.LineItem'],
-        RecursiveHierarchyQualifier: 'GenreHierarchy'
+        RecursiveHierarchyQualifier: 'GenreHierarchyHierarchy'
```
agoerler, line 107: I think we should avoid the duplicate HierarchyHierarchy

app/genres/webapp/manifest.json
```diff
@@ -93,7 +93,7 @@
 							"controlConfiguration": {
                                 "@com.sap.vocabularies.UI.v1.LineItem": {
                                     "tableSettings": {
-                                        "hierarchyQualifier": "GenreHierarchy",
+                                        "hierarchyQualifier": "GenreHierarchyHierarchy",
```
agoerler, line 96: I think we should avoid the duplicate HierarchyHierarchy

package.json
```diff
@@ -4,7 +4,8 @@
 	"description": "Generated by cds-services-archetype",
 	"devDependencies": {
 		"@sap/cds-dk": "9.9.1",
-		"@sap/cds-mtxs": "^3"
+		"@sap/cds-mtxs": "^3",
+		"@sap/cds-compiler": "^6.9.0"
```
beckermarc, line 8: Why is this required? I think we should avoid that explicit dependency in package.json

app/admin/fiori-service.cds
beckermarc: +1 I am also wondering: Why do we need an explicit entity `GenreHierarchy` at all? Why isn't `Genres` itself annotated with `@hierarchy`?

package.json
agoerler: Without this dependency we get cds-compiler 6.6. It appears that cds-dk 9.9 contains compiler 6.6 instead of 6.9.

app/admin/fiori-service.cds
eugene-andreev: Because the `@hierarchy` doesn't allow to specify the `ChangeNextSiblingAction: 'AdminService.moveSibling'`, thus this should be done in an old-style

package.json
beckermarc: I doubt this is the case. I think it would be good to investigate this further, this definitely shouldn't be required.

PR #602, opened 2026-01-28
CAP Java 5 und Spring boot 4 migration

StefanHenke:
Requires a Spring Boot 4 compatible version of CAP Java and the BTP Security Library which are not yet available on Maven Central.


mofterdinger:
@StefanHenke  Hi, I've updated this PR to CAP Java 5 and I bumped also some other dependencies. The update to cds-dk 10 ist still missing. Once this is done we should merge this PR.

srv/pom.xml
```diff
@@ -122,14 +127,26 @@
     </dependency>
 
     <dependency>
-      <groupId>org.springframework.security</groupId>
-      <artifactId>spring-security-test</artifactId>
+      <groupId>org.springframework.boot</groupId>
```
mofterdinger, line 130: are these new dependencies really required ?

mofterdinger (approved):
lets merge and do an update of the cds-dk once the next major is available. CAP Java 5 also supports the previous cds-dk 9.9.x

vmikhailenko (approved):
LGTM

PR #653, opened 2026-04-28
Postgres: Bump approuter version and increase disk-quota

mofterdinger:


vmikhailenko (approved):

mta.yaml
```diff
@@ -12,7 +12,7 @@ modules:
     path: srv
     parameters:
       memory: 1024M
-      disk-quota: 512M
+      disk-quota: 768M
```
vmikhailenko, line 15: They grow so fast. :)

PR #659, opened 2026-04-30
chore: Configure 'cds-feature-ord'

mlakov:


srv/src/main/resources/application.yaml
```diff
@@ -53,7 +53,10 @@ cds:
       kind: enterprise-messaging
       format: cloudevents
       subscribe-prefix: sap/S4HANAOD/java/ce/
-
+  ucl:
```
StefanHenke: Usage of certificates makes it very hard for consumers to locally test (which is typically what people do with the bookshop sample). One would need to integrate with UCL etc. As we don´t have basic auth available as of today for ord in java, I´m thinking of setting the switch to enable public metadata endpoints which is honoured by the ord feature

StefanHenke (commented):
The change as such look fine to me except one comment. However, the `cds-feature-ord` is currently not released externally which means we cannot merge it to `main`

mlakov:
> The change as such look fine to me except one comment. However, the `cds-feature-ord` is currently not released externally which means we cannot merge it to `main`

Are you sure that the `cds-feature-ord` is not released externally ? It is visible in [Maven Central](https://central.sonatype.com/artifact/com.sap.cds/cds-feature-ord) ?

srv/src/main/resources/application.yaml
mlakov: Resolved.

.cdsrc.json
```diff
@@ -2,7 +2,15 @@
...
         "multitenancy": true,
         "extensibility": true,
-        "toggles": true
+        "toggles": true,
+        "ord": true
     },
-    "profile": "with-mtx-sidecar"
+    "profile": "with-mtx-sidecar",
+    "ord": {
+        "namespace": "sap.sample",
```
mlakov, line 10: Add config property `packageName` in here

PR #682, opened 2026-06-15
Show User Description

agoerler:


srv/src/main/java/my/bookshop/handlers/UserDescriptionHandler.java
```diff
@@ -0,0 +1,31 @@
...
+public class UserDescriptionHandler implements EventHandler {
+
+  @On(service = "*")
+  protected void addDraftFields(DraftCreateEventContext context, List<CdsData> entries) {
+    CdsDataProcessor.Filter filter =
+        (path, element, type) ->
+            switch (element.getName()) {
+              case "InProcessByUserDescription" -> true;
+              default -> false;
+            };
```
beckermarc: Could also be expressed simpler with `"InProcessByUserDescription".equals(element.getName())` in my opinion

srv/src/main/java/my/bookshop/handlers/UserDescriptionHandler.java
```diff
@@ -0,0 +1,31 @@
...
+import com.sap.cds.services.handler.EventHandler;
+import com.sap.cds.services.handler.annotations.On;
+import com.sap.cds.services.request.UserInfo;
+import java.util.List;
+import org.springframework.stereotype.Component;
+
+@Component
+public class UserDescriptionHandler implements EventHandler {
+
+  @On(service = "*")
```
beckermarc: service * without type (here: DraftService) should be avoided. This handler is registered **everywhere**. I'd prefer `ServiceName` annotation on class-level.

beckermarc: Also: This should be a `@Before` handler, not an `@On` handler.

srv/src/main/resources/application.yaml
```diff
@@ -17,11 +17,21 @@ cds:
         password: admin
         roles:
         - admin
+        additional:
+            nickname: Chief Books Administrator
```
beckermarc: This is a little artificial. Would be more interesting to give an example that fetches the correct firstname and lastname from the XSUAA token (or IAS, if bookshop is meanwhile migrated to that) and then mocks these attributes in mock-user mode. That would be more valuable / realistic I think, also in case of deploying the bookshop to the cloud.

package.json
```diff
@@ -9,5 +9,10 @@
...
 	"workspaces": [
 		"mtx/sidecar"
-	]
-}
+	],
+	"cds": {
+		"cdsc": {
+			"draftUserDescription": true
+		}
+	}
```
beckermarc: Rather put this into `.cdsrc.json` where also the other `cds` configuration resides.

srv/src/main/resources/application.yaml
```diff
@@ -17,11 +17,23 @@ cds:
...
+        password: minad
+        roles:
+        - admin
+        additional:
+            given_name: Emma
+            family_name: Maier
       user:
         password: user
+        additional:
+            nickname: Some user
```
beckermarc:
```suggestion
```

srv/src/main/resources/application.yaml
```diff
@@ -17,11 +17,23 @@ cds:
...
         attributes:
           businessPartner:
           - "10401010"
+      minad:
+        password: minad
+        roles:
+        - admin
+        additional:
+            given_name: Emma
+            family_name: Maier
```
beckermarc: I think this would also need a `businessPartner` attributes, otherwise it will not be able to use the Manage Orders UI. However I think we can also omit that additional user.

srv/src/main/java/my/bookshop/handlers/UserDescriptionHandler.java
```diff
@@ -0,0 +1,36 @@
...
+  @Autowired private UserInfo userInfo;
+
+  @Before
+  protected void addDraftFields(DraftCreateEventContext context, List<CdsData> entries) {
+    CdsDataProcessor.Filter filter =
+        (path, element, type) -> "InProcessByUserDescription".equals(element.getName());
+    CdsDataProcessor.Generator generator = (path, element, isNull) -> description();
+    CdsDataProcessor dataProcessor = CdsDataProcessor.create().addGenerator(filter, generator);
+
+    dataProcessor.process(entries, context.getTarget());
```
beckermarc:
```suggestion
    CdsDataProcessor.create()
        .addGenerator(filter, generator)
        .process(entries, context.getTarget());
```

srv/src/main/resources/application.yaml
```diff
@@ -17,9 +17,22 @@ cds:
...
         roles:
         - admin
+        additional:
+            given_name: Paul
+            family_name: Müller 
         attributes:
           businessPartner:
           - "10401010"
+      minad:
+        password: minad
```
beckermarc:
```suggestion
      admin2:
        password: admin2
```

beckermarc (approved):

StefanHenke (approved):

